PR #4, opened 2018-11-14
BACKLOG-9012 Fix filter (for ancestors) when EQUAL is used so that it…

AKarmanov:
We need this change to be able to get ancestors of a node for "Locate" functionality in Content Manager.


tdraier:
The fix is not needed, you can access the field value by using "type.value" in the query : 
```parents:ancestors(fieldFilter: {filters: {fieldName: "type.value", evaluation: AMONG, values:["jnt:page", "jnt:folder", "jnt:contentFolder"]}})```

PR #8, opened 2019-04-23
BACKLOG-10130 add GraphQL field "displayName"

Taybou:


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/nodetype/GqlJcrPropertyDefinition.java
```diff
@@ -135,4 +137,12 @@ public GqlJcrPropertyType getRequiredType() {
         return GqlJcrPropertyType.fromValue(definition.getRequiredType());
     }
 
+    @GraphQLField
+    @GraphQLName("displayName")
+    @GraphQLNonNull
+    public String getDispalyName(@GraphQLName("language") @GraphQLNonNull String language) {
```
cedmail: Typo error in the method name, could you rename to getDisplayName ?

Taybou: Ah nice catch, thank you @cedmail

PR #10, opened 2019-06-12
BACKLOG-10433/MOD-1881

cedmail:
Branch used for developing new functionalities for the SDL API, allowing to declare custom fetchers, like this example from cloudinary-integration module

extend type ImageAsset {
    cloudinary: CloudinaryAsset
}

"""
Addon for Cloudinary integration
"""
type CloudinaryAsset {
    """Public identifier for Cloudinary account (null if Cloudinary settings is not configured)"""
    cloudspace: String @fetcher(name: "cloudinaryFetcher")
    """Path of the asset in Cloudinary (null if Cloudinary settings is not configured)"""
    path: String @fetcher(name: "cloudinaryFetcher")
    """Upload url for this asset (null if Cloudinary settings is not configured)"""
    url: String @fetcher(name: "cloudinaryFetcher")
}

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/sdl/parsing/SDLTypeChecker.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/sdl/parsing/SDLTypeChecker.java
@@ -9,6 +9,7 @@ import org.jahia.data.templates.JahiaTemplatesPackage;
 import org.jahia.modules.graphql.provider.dxm.sdl.SDLConstants;
 import org.jahia.modules.graphql.provider.dxm.sdl.parsing.status.SDLDefinitionStatus;
 import org.jahia.modules.graphql.provider.dxm.sdl.parsing.status.SDLDefinitionStatusType;
+import org.jahia.osgi.BundleUtils;
 import org.jahia.services.content.nodetypes.ExtendedNodeType;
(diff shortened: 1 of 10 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/sdl/parsing/SDLTypeChecker.java
```diff
@@ -153,9 +155,27 @@ private static SDLDefinitionStatus checkForFieldsConsistency(ObjectTypeDefinitio
                     l.add(fieldDefinition.getType().toString());
                 }
             });
+            //Verify that type exists in definition registry
             if (!l.isEmpty()) {
                 return new SDLDefinitionStatus(objectTypeDefinition.getName(), SDLDefinitionStatusType.MISSING_TYPE, StringUtils.join(l, ','));
             }
+            //Validate that if custom fetchers are declared (through fetcher directive) they exist in the map of registered fetchers.
+            SDLSchemaService service = BundleUtils.getOsgiService(SDLSchemaService.class, null);
```
tdraier: Would it not be better to pass the SDLSchemaService as a parameter instead of looking up viaBundleUtils ?

ssavu: Good point, I have just updated the code to reflect your suggestion.

PR #12, opened 2019-07-15
QA-11925 add GraphQL fields formattedDateValue and formattedDateValues

Taybou:


dgriffon:
Thank you for the pull request !
As said in commented code, can you use the date format provide by the definition to send and save the date. Here sample property definitions with that option (from QA Modules):
```
 - mixin4DateCustom (date, datepicker[format='yyyy-MM-dd']) 
 - mixin4DatetimeCustom (date, datetimepicker[format='yyyy-MM-dd HH:mm'])
```
Can you add a unit test using date format on the new fields ? 
Thank you again !

PR #16, opened 2019-09-13
BACKLOG-11050 : add useAvailableNodeName parameter to addNode() to al…

dgriffon:
…low existing name to be suffixed when necessary.


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeInput.java
```diff
@@ -56,17 +56,20 @@
...
     private String name;
     private String primaryNodeType;
+    private Boolean useAvailableNodeName;
     private Collection<String> mixins;
     private Collection<GqlJcrPropertyInput> properties;
     private Collection<GqlJcrNodeInput> children;
 
     public GqlJcrNodeInput(@GraphQLName("name") @GraphQLNonNull @GraphQLDescription("The name of the node to create") String name,
-                           @GraphQLName("primaryNodeType") @GraphQLNonNull @GraphQLDescription("The primary node type of the node to create") String primaryNodeType,
+                           @GraphQLName("primaryNodeType") @GraphQLDescription("The primary node type of the node to create") String primaryNodeType,
```
tdraier: Why not @GraphQLNonNull anymore ?

dgriffon: it was a typo, thanks for notify it!
fixed by amending initial commit

PR #21, opened 2019-10-17
BACKLOG-11362: add isDisplayableNode end point in RenderNodeExtensions

jkevan:


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/render/RenderNodeExtensions.java
```diff
@@ -103,6 +107,25 @@ public GqlJcrNode getDisplayableNode(DataFetchingEnvironment environment) {
         }
     }
 
+    /**
+     * Check if the node as a renderable template associated with it (not a view a template).
```
dgriffon:
```suggestion
     * Check if the node has a renderable template associated with it (not a view a template).
```

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/render/RenderNodeExtensions.java
```diff
@@ -103,6 +107,25 @@ public GqlJcrNode getDisplayableNode(DataFetchingEnvironment environment) {
...
+     *
+     * @return true if the node as a renderable template associated with it
```
dgriffon:
```suggestion
     * @return true if the node has a renderable template associated with it
```

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/render/RenderNodeExtensions.java
```diff
@@ -103,6 +107,25 @@ public GqlJcrNode getDisplayableNode(DataFetchingEnvironment environment) {
...
+     */
+    @GraphQLField
+    @GraphQLName("isDisplayableNode")
+    @GraphQLDescription("Check if the node as a renderable template associated with it (not a view a template).")
```
dgriffon:
```suggestion
    @GraphQLDescription("Check if the node has a renderable template associated with it (not a view a template).")
```

PR #26, opened 2020-02-10
BACKLOG-12303 : add tag suggestion endpoint

jsinovassin:
https://jira.jahia.org/browse/BACKLOG-12303

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/GqlTagQuery.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/GqlTagQuery.java
@@ -1,0 +1,114 @@
+/*
+ * ==========================================================================================
+ * =                   JAHIA'S DUAL LICENSING - IMPORTANT INFORMATION                       =
+ * ==========================================================================================
+ *
+ *                                 http://www.jahia.com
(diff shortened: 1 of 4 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/GqlTagQuery.java
```diff
@@ -0,0 +1,118 @@
...
+            @GraphQLName("prefix") @GraphQLNonNull @GraphQLDescription("The prefix of the tags") String prefix,
+            @GraphQLName("limit") @GraphQLNonNull @GraphQLDescription("Max number of tags to get") Long limit,
+            @GraphQLName("startPath") @GraphQLNonNull @GraphQLDescription("The root node to start the search") String startPath,
+            DataFetchingEnvironment environment
+    ) throws RepositoryException {
+
+        try {
+            JCRSessionWrapper jcrSessionWrapper = getSession();
+
+            TaggingService taggingService = BundleUtils.getOsgiService(TaggingService.class, null);
```
jkevan: Could be great to map all the properties hardcoded like: mincount, offset and sortByCount
so we can specified them on the graphQL call if necessary
they could be null, but it can be useful in some cases

PR #28, opened 2020-03-04
BACKLOG-12336: custom safe check if clearLock permission exist, becau…

jkevan:


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/locking/LockJCRNodeMutationExtension.java
```diff
@@ -124,7 +125,20 @@ public boolean unlock(@GraphQLName("type") @GraphQLDefaultValue(DefaultLockTypeP
...
-            if (nodeToUnlock.hasPermission("clearLock")) {
+
+            // custom check for permission added in DX 7.3.5.0:
+            boolean clearLockPermissionExist = false;
+            try {
+                // will crash in case the permission doesnt exist
+                nodeToUnlock.getAccessControlManager().privilegeFromName("clearLock");
+                // permission exist
+                clearLockPermissionExist = true;
+            } catch (RepositoryException e) {
```
cedricmoitrier:
```suggestion
            } catch (AccessControlException e) {
```

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/locking/LockJCRNodeMutationExtension.java
```diff
@@ -124,7 +125,20 @@ public boolean unlock(@GraphQLName("type") @GraphQLDefaultValue(DefaultLockTypeP
     public boolean clearAllLocks() {
         try {
             JCRNodeWrapper nodeToUnlock = nodeMutation.getNode().getNode();
-            if (nodeToUnlock.hasPermission("clearLock")) {
+
+            // custom check for permission added in DX 7.3.5.0:
```
cedricmoitrier: I would extract the whole logic to check whether or not the user has the permission (including the fact that the permission exist or not) outside of this method (inside this class if fine).

PR #32, opened 2020-04-10
BACKLOG-12896 - Avoid exception log if testing a permission that doesn't exist.

sergehuber:
This is a second attempt at implementing this properly since the first version caused problems with system permission names.

## JIRA

https://jira.jahia.org/browse/BACKLOG-12896

## Description

This change is needed by the new Jahia product home page because the documentation cards might use permissions that are not deployed on some installations, such as the jExperience global access permission when jExperience is not yet deployed. Before this change it would log an exception coming from the JCRNodeWrapperImpl.hasPermission implementation which could lead users to think that something is seriously wrong when it is just a permission check on an non-declared permission which should just fail.

This change implements this at the GraphQL level to avoid changing the underlying logic for other users and implementations.

## Tests

- [ ] Unit Tests (Most changes _should_ have unit tests)
- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


sergehuber:
Thanks Thomas for the review

PR #33, opened 2020-04-14
QA-12458 improve get ref node

Taybou:
## JIRA

https://jira.jahia.org/browse/QA-12458

## Description
Improve get ref node

## Tests

- [ ] Unit Tests (Most changes _should_ have unit tests)
- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrProperty.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrProperty.java
@@ -341,7 +341,7 @@ public class GqlJcrProperty {
     @GraphQLDescription("GraphQL representation of the node this property references in case the property is single-valued, null otherwise")
     public GqlJcrNode getRefNode() throws GqlJcrUnresolvedNodeReferenceException {
         try {
-            if (property.isMultiple() || property.getType() != PropertyType.REFERENCE || property.getType() != PropertyType.WEAKREFERENCE) {
+            if (property.isMultiple() || (property.getType() != PropertyType.REFERENCE && property.getType() != PropertyType.STRING)) {
                 return null;
             }
             return getRefNode(property.getValue());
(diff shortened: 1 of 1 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrProperty.java
```diff
@@ -341,7 +341,7 @@ public Double getFloatValue() {
     @GraphQLDescription("GraphQL representation of the node this property references in case the property is single-valued, null otherwise")
     public GqlJcrNode getRefNode() throws GqlJcrUnresolvedNodeReferenceException {
         try {
-            if (property.isMultiple()) {
+            if (property.isMultiple() || property.getType() != PropertyType.REFERENCE || property.getType() != PropertyType.WEAKREFERENCE) {
```
tdraier, line 344: In which case could `(property.getType() != PropertyType.REFERENCE || property.getType() != PropertyType.WEAKREFERENCE)` ever be false ?

Taybou: no idea, it's what was suggested in the ticket.

Taybou: And when I tested in the node properties:
- the j:nodename for example it's of type reference

tdraier: Hint : it's not possible to have property.getType() (a number) equals to PropertyType.REFERENCE and PropertyType.WEAKREFERENCE at the same time. So the opposite condition will always be true, hence the condition is broken (can be simplified to if (true) ) and you will never get a value anymore when calling getRefNode

PR #52, opened 2020-08-28
BACKLOG-14531 : add graphQL playground into developer tools

Yousria:


graphql-dxm-provider/src/main/resources/tools/graphql-playground.jsp
```diff
@@ -0,0 +1,57 @@
...
+        <title>GraphQL Playground</title>
+        <link rel="stylesheet" href="${modulePath}/css/graphql-playground.css" type="text/css"/>
+    </head>
+    <body>
+        <div id="graphql-playground">
+            <p class="loading">Loading <span class="title">GraphQL Playground</span></p>
+        </div>
+        <script src="${modulePath}/javascript/graphql-playground-1.7.23.js"></script>
+        <script>
+            const customTheme = {
```
cedricmoitrier, line 20: You can remove this customTheme, I was playing with colors.

graphql-dxm-provider/src/main/resources/tools/graphql-playground.jsp
```diff
@@ -0,0 +1,57 @@
...
+            };
+
+            window.addEventListener('load', function() {
+                GraphQLPlayground.init(document.getElementById('graphql-playground'), {
+                    endpoint: '${graphQlEndpoint}',
+                    settings: {
+                        'request.credentials': 'same-origin'
+                    },
+                    shareEnabled: true,
+                    //workspaceName: 'Jahia',
```
cedricmoitrier, line 51: You can remove both commented attributes (workspaceName and codeTheme)

graphql-dxm-provider/src/main/resources/tools/graphql-playground.jsp
```diff
@@ -0,0 +1,57 @@
...
+                // rightDrawerBackground: ''
+            };
+
+            window.addEventListener('load', function() {
+                GraphQLPlayground.init(document.getElementById('graphql-playground'), {
+                    endpoint: '${graphQlEndpoint}',
+                    settings: {
+                        'request.credentials': 'same-origin'
+                    },
+                    shareEnabled: true,
```
cedricmoitrier, line 50: This one enables integration with https://www.graphqlbin.com/ to share a workspace. Don't know whether or not we should enable that. Should double check with @fpral

cedricmoitrier (commented):
I don't recall exactly why I did not use 1.8.x branch in the end (stability?). But a 1.7.26 has been released in between, thus we should use this one over the 1.7.23 one.

graphql-dxm-provider/src/main/resources/tools/graphql-playground.jsp
Yousria: > I don't recall exactly why I did not use 1.8.x branch in the end (stability?). But a 1.7.26 has been released in between, thus we should use this one over the 1.7.23 one.

I couldn't use 1.7.26 version tried to make it work but It doesn't load constantly, here is a screenshot
Capture d’écran 2020-09-01 à 09 07 55

cedricmoitrier (approved):

PR #61, opened 2020-10-19
QA-13171 : waiting until the publication is finished

jsinovassin:
https://jira.jahia.org/browse/QA-13171

### DIFF
--- a/graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
+++ b/graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
@@ -173,6 +173,7 @@ public class GraphQLPublicationTest extends GraphQLTestSupport {
 
         long startedWaitingAt = System.currentTimeMillis();
 
+        // Wait until the node is published via a background job.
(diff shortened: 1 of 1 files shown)


graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
```diff
@@ -166,28 +169,28 @@ private void testPublication(boolean publishSubNodes) throws RepositoryException
...
             if (System.currentTimeMillis() - startedWaitingAt > TIMEOUT_WAITING_FOR_PUBLICATION) {
                 Assert.fail("Timeout waiting for node to be published");
             }
+        }
+
+        JSONObject mutationResult = result.getJSONObject("data").getJSONObject("jcr").getJSONObject("mutateNode");
+        Assert.assertTrue(mutationResult.getBoolean("publish"));
+
+        // Wait until the node is published via a background job.
+        do {
```
dgriffon, line 186: You don't  need the do/while statement here as you already know that no publication job is running.

dgriffon (approved):

PR #66, opened 2020-10-30
Backlog 14944 implement stub version query for admin nodes

kovunov:
## JIRA
https://jira.jahia.org/browse/BACKLOG-14944

## Description
Added stub query and mutation for product version.


Fgerthoffert:
You'd need to branch off a fresh 1_x, the commit history on this branch is not clean (i.e. we shouldn't have the skip ci commits).

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlAdminQuery.java
```diff
@@ -1,12 +1,23 @@
...
  * GraphQL root object for Admin related queries.
  */
 @GraphQLName("adminQuery")
 @GraphQLDescription("Admin queries root")
 public class GqlAdminQuery {
+
+    @GraphQLField
+    @GraphQLName("version")
+    @GraphQLNonNull
+    @GraphQLDescription("Stub query which displays current Jahia product version")
```
Fgerthoffert, line 19: Just `Version of the running Jahia instance`, not need to say that's a "stub" we did for other reasons

PR #80, opened 2020-11-26
BACKLOG-13977: Added new property to return the binary size for a node

nbugash-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-13977

## Description

Added a new property to return the binary size of the node

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrProperty.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrProperty.java
@@ -314,6 +314,23 @@ public class GqlJcrProperty {
     }
 
     /**
+     * @return The binary size of the JCR node as a Long, null otherwise
+     */
+    @GraphQLField
+    @GraphQLName("size")
+    @GraphQLDescription("The binary size of the JCR node as a Long, null otherwise")
+    public Long getSize() {
+        try {
+            if (property.isMultiple()) {
+                return null;
+            }
+            return property.getValue().getBinary().getSize();
+        } catch (RepositoryException e) {
+            throw new DataFetchingException(e);
+        }
+    }
+
+    /**
      * @return The values of the JCR property as Longs in case the property is multiple-valued, null otherwise
      */
     @GraphQLField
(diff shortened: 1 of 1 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrProperty.java
```diff
@@ -313,6 +313,23 @@ public Long getLongValue() {
         }
     }
 
+    /**
+     * @return The binary size of the JCR node as a Long, null otherwise
+     */
+    @GraphQLField
+    @GraphQLName("size")
+    @GraphQLDescription("The binary size of the JCR node as a Long, null otherwise")
+    public Long getSize() {
```
cedricmoitrier, line 322: Note quite sure this is the best approach to have this here - but that module is kind of unknown territory for me.
Thoughts @AKarmanov @tdraier

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrProperty.java
```diff
@@ -313,6 +313,23 @@ public Long getLongValue() {
...
+        try {
+            if (property.isMultiple()) {
```
cedricmoitrier, line 324: You should test the type of the property here as well and return null if not a BINARY.

cedricmoitrier (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrProperty.java
```diff
@@ -313,6 +313,23 @@ public Long getLongValue() {
...
 
+    /**
+     * @return The binary size of the JCR node as a Long, null otherwise
+     */
+    @GraphQLField
+    @GraphQLName("size")
+    @GraphQLDescription("The binary size of the JCR node as a Long, null otherwise")
+    public Long getSize() {
+        try {
+            if (property.isMultiple() && property.getValue().getType() != PropertyType.BINARY) {
```
cedricmoitrier, line 324:
```suggestion
            if (property.isMultiple() && property.getType() != PropertyType.BINARY) {
```

cedricmoitrier (approved):

PR #86, opened 2021-01-12
[BACKLOG-15213] Adapt the test following the deprecation of a node type

rknj:
## JIRA

https://jira.jahia.org/browse/BACKLOG-15213

### DIFF
--- a/graphql-test/src/main/java/org/jahia/test/graphql/GraphQLNodeTypesTest.java
+++ b/graphql-test/src/main/java/org/jahia/test/graphql/GraphQLNodeTypesTest.java
@@ -194,34 +194,34 @@ public class GraphQLNodeTypesTest extends GraphQLTestSupport {
 
     @Test
     public void shouldGetPropertyDefinitionWithConstraints() throws Exception {
-        JSONObject result = executeQuery("{\n" + 
-                "  jcr {\n" + 
-                "    nodeByPath(path: \"/j:acl/GRANT_g_users\") {\n" + 
-                "      uuid\n" + 
-                "      path\n" + 
-                "      name\n" + 
-                "      aceType: property(name: \"j:aceType\") {\n" + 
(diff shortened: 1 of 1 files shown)


graphql-test/src/main/java/org/jahia/test/graphql/GraphQLNodeTypesTest.java
```diff
@@ -410,6 +410,6 @@ public void shouldRetrieveIncludedAndNotExcludedNodeTypes() throws Exception {
             Assert.assertFalse(nt.isNodeType("jmix:hiddenType"));
         }
 
-        Assert.assertEquals(11, nodeTypes.length());
+        Assert.assertEquals(10, nodeTypes.length());
```
cedmail, line 413: no need to assert twice the same thing

graphql-test/src/main/java/org/jahia/test/graphql/GraphQLNodeTypesTest.java
```diff
@@ -401,7 +401,7 @@ public void shouldRetrieveIncludedAndNotExcludedNodeTypes() throws Exception {
         JSONArray nodeTypes = result.getJSONObject("data").getJSONObject("jcr").getJSONObject("nodeTypes").getJSONArray("nodes");
 
         NodeTypeRegistry nodeTypeRegistry = NodeTypeRegistry.getInstance();
-        Assert.assertEquals(11, nodeTypes.length());
+        Assert.assertEquals(10, nodeTypes.length());
```
cedmail, line 404: To improve stability just assert that there is results in the array, we found something.
Next loop make sure that what we found matches the criteria. 
How many results is not the matter in this test, otherwise we should ensure which 10 we found

cedmail (changes requested):

tdraier (approved):

PR #88, opened 2021-01-20
BACKLOG-15343: GraphQLPublicationTest failing randomly

nbugash-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-15343

## Description

Added some assertions and logs for more transparency regarding these failing tests

## Tests

- [x] Integration Tests

### DIFF
--- a/graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
+++ b/graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
@@ -54,12 +54,15 @@ import org.jahia.services.usermanager.JahiaGroupManagerService;
 import org.jahia.services.usermanager.JahiaUser;
 import org.jahia.services.usermanager.JahiaUserManagerService;
 import org.jahia.test.TestHelper;
+import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.junit.*;
 import org.junit.rules.TestName;
 import org.quartz.JobDetail;
 import org.quartz.SchedulerException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.util.CollectionUtils;
 
 import javax.jcr.PathNotFoundException;
@@ -71,7 +74,7 @@ public class GraphQLPublicationTest extends GraphQLTestSupport {
 
     @Rule public TestName name = new TestName();
 
-
+    private final Logger logger = LoggerFactory.getLogger(GraphQLPublicationTest.class);
     private static final long TIMEOUT_WAITING_FOR_PUBLICATION = 5000;
     private static final String TESTSITE_NAME = "graphqlPublicationTestSite";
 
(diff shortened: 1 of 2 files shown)


graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
```diff
@@ -183,11 +186,27 @@ private void testPublication(boolean publishSubNodes) throws RepositoryException
...
+                + "         property(name: \"j:published\") {"
+                + "            value"
+                + "         }"
+                +"      }"
+                +"   }"
+                +"}"
+        );
+        JSONObject queryResult = expectedResult.getJSONObject("data").getJSONObject("jcr").getJSONObject("nodeByPath").getJSONObject(
+                "property");
+        Assert.assertTrue("The node: '/sites/" + siteName + "/testList' should have been published", queryResult.getBoolean("value"));
```
tdraier, line 202: The assert will stop the execution if it fails and won't give more information (it won't go into debugSession). You can also maybe execute the same query in default (since it's quite clear the node does not exist in live when it fails)

nbugash-jahia: I see. I'll move it because I thought it'll continue with the test even if the assert at line 202 fails

graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
```diff
@@ -296,4 +299,81 @@ private JCRSiteNode createTestSite(String name, String template, Set<String> lan
...
+        // Method #1
+        long startedWaitingAt = System.currentTimeMillis();
+
+        // Wait until the node is published via a background job.
+        while(Arrays.stream(schedulerService.getScheduler().getTriggerNames(SchedulerService.INSTANT_TRIGGER_GROUP))
+                .anyMatch(t -> t.contains("Publication"))) {
+            if (System.currentTimeMillis() - startedWaitingAt > TIMEOUT_WAITING_FOR_PUBLICATION) {
+                Assert.fail("Timeout waiting for node to be published");
+            }
+        }
```
bpapez, line 363: Is it really enough to check for the triggerNames? Don't we have to add the getAllActiveJobs afterwards? You could set a breakpoint in the PublicationJob to artificially delay/stop it and see that this check really waits until the job is finished.

graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
```diff
@@ -296,4 +299,81 @@ private JCRSiteNode createTestSite(String name, String template, Set<String> lan
...
+                .ignoreExceptions()
+                .until(new Callable<Boolean>() {
+                    @Override public Boolean call() throws Exception {
+                        final String[] triggerNames = schedulerService.getScheduler()
+                                .getTriggerNames(SchedulerService.INSTANT_TRIGGER_GROUP);
+                        return Arrays.stream(triggerNames).noneMatch(t -> t.contains("Publication"));
+                    }
+                }, equalTo(true));
+         */
+    }
```
bpapez, line 378: To make awaitility work you probably need to add the library to the pom.xml so it gets packaged with the module

bpapez (changes requested):

graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
nbugash-jahia: i see

graphql-test/src/main/java/org/jahia/test/graphql/GraphQLPublicationTest.java
```diff
@@ -296,4 +296,22 @@ private JCRSiteNode createTestSite(String name, String template, Set<String> lan
...
+                    final boolean publicationJobs = allActiveJobs.stream().anyMatch(job -> job.getDescription().equals("Publication"));
+                    logger.debug("Trigger names: {}", Arrays.toString(triggerNames));
+                    allActiveJobs.forEach(j -> logger.debug("Active job '{}'", j.getDescription()));
+                    return Arrays.stream(triggerNames).anyMatch(triggerName -> triggerName.contains("Publication")) && publicationJobs;
+                }
+            }, equalTo(false));
+        } catch (ConditionTimeoutException e) {
+            Assert.fail("Publication job did not finish within the set timeout of " + TIMEOUT_WAITING_FOR_PUBLICATION + " ms");
+        }
+    }
```
bpapez, line 316: For me the method is difficult to read and understand what it is doing. Also it is not necessary to check for triggers and active jobs in each poll iteration. I would rather suggest something like that:

```
    private void waitForPublicationToFinish() {
        SchedulerService schedulerService = BundleUtils.getOsgiService(SchedulerService.class, null);
        ConditionFactory conditionalWait = with().pollInterval(ONE_SECOND).await().atMost(TIMEOUT_WAITING_FOR_PUBLICATION, MILLISECONDS);
        // wait until no more PublicationJob triggers are scheduled 
        conditionalWait.until(() -> Arrays.stream(schedulerService.getScheduler().getTriggerNames(SchedulerService.INSTANT_TRIGGER_GROUP))
                .noneMatch(triggerName -> triggerName.contains(PUBLICATION_JOB_NAME)));
        // wait until last PublicationJob is finished
        conditionalWait.until(() -> schedulerService.getAllActiveJobs().stream()
                .noneMatch(job -> job.getJobClass().getName().contains(PUBLICATION_JOB_NAME)));
    }
```

The added logging and the exeption catch/assertion IMHO are not necessary.

bpapez (changes requested):

bpapez (approved):

PR #107, opened 2021-04-13
Copy report from the test run at compilation

rknj:
## Description

Some tests are run during the build and we need to add the report in the tests tab (with all the generated artifacts)


Fgerthoffert:
Just wondering, will CircleCI know to fail the build if there's an error in one of the xml files ?

Fgerthoffert (approved):

rknj:
> Just wondering, will CircleCI know to fail the build if there's an error in one of the xml files ?

Normally the build will fail if there's an error in these tests. I just wanted to make the report available  in CircleCI.

PR #117, opened 2021-05-20
QA-13720 updating pom xml.

XiangYuHe:
## JIRA

https://jira.jahia.org/browse/QA-13720

## Description

Updating pom xml for the graph-ql to bring in the latest changes from jahia-impl.

### DIFF
--- a/graphql-dxm-provider/pom.xml
+++ b/graphql-dxm-provider/pom.xml
@@ -126,7 +126,7 @@
         <dependency>
             <groupId>org.jahia.test</groupId>
             <artifactId>module-test-framework</artifactId>
-            <version>8.0.0.0</version>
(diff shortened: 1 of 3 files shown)


tdraier (changes requested):
Can you be more specific on the reason which forces you to change dependency and break compatibility ? I don't see any modification in the code which requires such a change.

XiangYuHe:
@tdraier the reason I am changing the dependency is due to I am updating the jahia-impl java code.
The graphql-core need to get that latest code for this change.

tdraier:
I don't see why the check on the property is required for graphql provider. And I don't see any change in graphql which would require a new dependency. There's a  bug in the core, not in graphql-provider.

PR #128, opened 2021-06-18
BACKLOG-16199 prepare graphql-core 1_x for release

kovunov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16199

## Description
Prepare graphql-core `1_x` for release

### DIFF
--- a/.circleci/.circleci.settings.xml
+++ b/.circleci/.circleci.settings.xml
@@ -51,5 +51,13 @@
             <username>${env.NEXUS_USERNAME}</username>
             <password>${env.NEXUS_PASSWORD}</password>
         </server>
+        <server>
+            <id>staging-repository</id>
+            <username>${env.NEXUS_RELEASE_USERNAME}</username>
+            <password>${env.NEXUS_RELEASE_PASSWORD}</password>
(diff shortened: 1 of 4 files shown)


Fgerthoffert:
Could you use the orb instead of declaring the job directly in the config file?

There might be a valid reason not to, but if the job is the same, then we should try to use the orb.

kovunov:
> Could you use the orb instead of declaring the job directly in the config file?
> 
> There might be a valid reason not to, but if the job is the same, then we should try to use the orb.
> 
> https://github.com/Jahia/jahia-modules-orb/blob/main/src/jobs/release.yml

Orb for instance uses `yarn` and `yarn auto` in the release job, while in the graphql-dxm-provider we don't need it. We can switch to orb, but I don't see it as absolutely necessary on this branch.

Fgerthoffert:
I think we should update the orb to use npx or something, it's not the first time I'm hearing this comment :)

The only reason we have yarn here is to use auto and generate the changelog, it's not realted to the project having/being javascript.

Fgerthoffert (approved):

PR #129, opened 2021-06-21
Use the newest Jahia images for the tests

rknj:


.circleci/config.yml
```diff
@@ -495,7 +495,7 @@ workflows:
             - jahia-modules-orb/build
           context: QA_ENVIRONMENT
           #Using Aliases here (8 is the latest of the 8 tree, 8.1 is the latest of the tree, ...)
-          JAHIA_IMAGE: jahia/jahia-private:snapshot-dev
+          JAHIA_IMAGE: jahia/jahia-ee-dev:8-SNAPSHOT
```
tdraier, line 498: shouldn't we also update the docker-compose file ? the parameters of old and new image are completely different. "TOOLS_USER" and "TOOLS_PWD" are not used at all

Fgerthoffert (approved):

PR #136, opened 2021-06-23
BACKLOG-16194 expand user type

kovunov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16194

## Description

Expanded user type and added some tests that properties are present

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/user/GqlUser.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/user/GqlUser.java
@@ -43,10 +43,7 @@
  */
 package org.jahia.modules.graphql.provider.dxm.user;
 
-import graphql.annotations.annotationTypes.GraphQLDescription;
-import graphql.annotations.annotationTypes.GraphQLField;
-import graphql.annotations.annotationTypes.GraphQLName;
-import graphql.annotations.annotationTypes.GraphQLNonNull;
+import graphql.annotations.annotationTypes.*;
 import graphql.annotations.connection.GraphQLConnection;
 import graphql.schema.DataFetchingEnvironment;
 import org.jahia.data.viewhelper.principal.PrincipalViewHelper;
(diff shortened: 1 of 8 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/user/GqlUser.java
```diff
@@ -86,11 +83,56 @@ public GqlUser(JahiaUser jahiaUser) {
 
     @GraphQLField
     @GraphQLNonNull
+    @GraphQLDeprecate
```
tdraier, line 86: you can remove the field if needed, it has never been released anyway :-)

tests/cypress/integration/api/admin/user.spec.ts
```diff
@@ -1,27 +1,32 @@
...
-                        admin {
-                            userAdmin {
-                                user(userName: "root") {
-                                    name
-                                }
-                            }
-                        }
-                    }
-                `,
+                query: GQL_USER,
```
tdraier, line 17: maybe you could use the GQL_USER query in other tests  ?

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/user/GqlUser.java
kovunov: This User object is also used by `currentUser` query so I didn't remove it in case the current user is used elsewhere, if it's safe from your point of view I can remove it

tests/cypress/integration/api/admin/user.spec.ts
kovunov: will do it now

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/user/GqlUser.java
tdraier: i did not realize it was already there, so let's keep it

tdraier (approved):

PR #137, opened 2021-06-23
[Test] - Testing Markdown in the GraphQL description

Fgerthoffert:
Desc in the title, do not merge


Fgerthoffert:
Screen Shot 2021-06-23 at 12 31 40 PM (1)

Fgerthoffert:
See screenshot about for an example of how this would look in GraphQL playground

PR #139, opened 2021-06-28
Update step waiting for Jahia to be alive

rknj:


Fgerthoffert (approved):

Fgerthoffert:
Will be merging since those changes are not associated with the failure on the other integration tests

PR #159, opened 2021-08-10
TECH-210: Fix graphql file mutation test

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/TECH-210

## Description

Fix GraphQL mutation upload test

I've explored multiple ways of fixing the tests. One is the current way of injecting the file to test into the GraphQL context. Fixed by creating a custom GraphQL context and ContextBuilder to be able to inject into the service, and wrap old FileItem into a javax.http.servlet.Part instance.

I've also tried another route of creating a Multipart request that attaches file contents but had some issue with having both file content and the query request (it wasn't obvious how both are attached and probably needs further exploration later). I've kept the code as separate functions but not used just in case we want to continue exploring that path later on.

### DIFF
--- a/graphql-test/src/main/java/org/jahia/test/graphql/GraphQLNodeMutationsTest.java
+++ b/graphql-test/src/main/java/org/jahia/test/graphql/GraphQLNodeMutationsTest.java
@@ -59,7 +59,9 @@ import org.junit.*;
 
 import javax.jcr.PropertyType;
 import javax.jcr.RepositoryException;
+import javax.servlet.http.Part;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.text.SimpleDateFormat;
 import java.util.*;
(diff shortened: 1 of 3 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/upload/UploadHelper.java
```diff
@@ -131,7 +131,7 @@ private static Part getPartForFilename(GraphQLServletContext context, String fil
 
     // https://docs.oracle.com/javaee/6/tutorial/doc/glraq.html
     private static String getFileName(final Part part) {
-        final String partHeader = part.getHeader("content-disposition");
+        final String partHeader = part.getHeader("Content-Disposition");
```
gflores-jahia, line 134: Couldn't find a constant to use for this header

tdraier (approved):

PR #162, opened 2021-08-13
BACKLOG-16328: Fixed error handling, add errorType in return

tdraier:
Servlet is now using a dedicated method to serialize the error, instead of jackson.
A new ErrorClassification has been added and is more fexible than the previous error type, which allows us to directly return the value we want in `getErrorType` and get rid of `getRealErrorType`

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLError.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLError.java
@@ -43,9 +43,7 @@
  */
 package org.jahia.modules.graphql.provider.dxm;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
+import graphql.ErrorClassification;
 import graphql.ErrorType;
 import graphql.GraphQLError;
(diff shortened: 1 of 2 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLError.java
```diff
@@ -84,23 +82,36 @@ public String getMessage() {
...
-    public ErrorType getErrorType() {
-        return errorType;
-    }
-
-    @JsonProperty("errorType")
-    public String getRealErrorType() {
+    public ErrorClassification getErrorType() {
         if (errorType != null) {
-            return errorType.toString();
+            return errorType;
```
gflores-jahia, line 87: I think we need to wrap errorType here to a SimpleClassification object as well?

gflores-jahia (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLError.java
tdraier: > I think we need to wrap errorType here to a SimpleClassification object as well? errorType is declared as an ErrorType enum object

ErrorType is already an ErrorClassification

gflores-jahia (approved):

PR #164, opened 2021-08-16
TECH-468: Upgrade to v16 (WIP)

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/TECH-468

## Description

* Added GqlTypeUtil to handle extracting/unwrapping`GraphQLType`s
* Refactored `GraphQLType.getName()` references

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
@@ -284,6 +284,14 @@ public class DXGraphQLProvider implements GraphQLTypesProvider, GraphQLQueryProv
     }
 
     @Override
+    public Collection<GraphQLFieldDefinition> getFields() {
+        /*
+         * FIXME: part of interface for both mutation and subscription
+         */
+        return null;
+    }
+
+    @Override
     public GraphQLCodeRegistry getCodeRegistry() {
         return codeRegistry;
     }
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
@@ -60,6 +60,8 @@ import org.jahia.modules.graphql.provider.dxm.relay.DXConnection;
 import org.jahia.modules.graphql.provider.dxm.relay.DXRelay;
 import org.jahia.modules.graphql.provider.dxm.sdl.parsing.SDLSchemaService;
 import org.jahia.modules.graphql.provider.dxm.security.GraphQLFieldWithPermissionRetriever;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.component.annotations.*;
 import org.reactivestreams.Publisher;
 import org.slf4j.Logger;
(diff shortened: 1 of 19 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
```diff
@@ -283,6 +283,14 @@ public void activate() {
         return subscriptionType.getFieldDefinitions();
     }
 
+    @Override
+    public Collection<GraphQLFieldDefinition> getFields() {
+        /*
+         * FIXME: part of interface for both mutation and subscription
+         */
+        return null;
```
gflores-jahia, line 291: @tdraier this one not sure how to resolve. the mutation and subscription provider interfaces has the same method. Is there a way to tell if it's a mutation or subscription provider? Or maybe we should rename `getMutationFields` and `getSubscriptionFields` ?

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/predicate/FieldEvaluator.java
```diff
@@ -146,21 +142,12 @@ public static FieldEvaluator forConnection(DataFetchingEnvironment environment)
...
-                } else if (fields.containsKey("edges")) {
-                    // If no "nodes" was found, try to look into { edges { node } } selection set
-                    MergedField edgeFields = fields.get("edges");
-                    fields = fieldCollector.collectFields(parameters, edgeFields).getSubFields();
-                    if (fields.containsKey("node")) {
-                        return getField(fieldCollector.collectFields(parameters, fields.get("node")), name);
-                    }
+                DataFetchingFieldSelectionSet selectionSet = environment.getSelectionSet();
+                if (selectionSet.contains("nodes")) {
+                    return getField(selectionSet.getFieldsGroupedByResultKey("nodes/**"), name);
```
gflores-jahia, line 148: It seems that retrieving fields from a selectionSet now supports glob pattern and supports recursing down fields and subfields, so not sure if fieldCollector is still needed here. These are all still untested (will try them out tomorrow) so not sure how valid of a solution this is yet.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/predicate/FieldEvaluator.java
```diff
@@ -240,22 +229,12 @@ public Object getFieldValue(Object source, String fieldName) {
                 .executionId(environment.getExecutionId())
                 .fragmentsByName(environment.getFragmentsByName())
                 .graphQLSchema(environment.getGraphQLSchema())
+                .arguments(environment.getArguments())
```
gflores-jahia, line 232: I don't know if this is a valid substitute to `valuesResolver.getArgumentValues` but don't know if there's a way to get to a `Field` object from `SelectedField` or maybe `DataFetchingFieldSelectionSet`

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
tdraier: getFields is used in servlet, for both mutation and subscription - the only way I see is to split the providers for these 2 interfaces

tdraier:
I have a lot of different issues with these new versions : 

- graphql-java 16 does not support osgi anymore. When upgrading grade, they had an issue with osgi plugin, and so just removed it. It seems it has been reintroduced in 17 ( https://github.com/graphql-java/graphql-java/commit/dbbe18abb0531b67abf825b014fa02e5870f9e74 )
- graphq-java-annotation 8.5 also has an issue with osgi - it has the plugin, but does not export any package (?)
- graphql-java-servlet 11.1.0 is ok, but it's not the last version - 11.1.1 is completely incompatible because of a dependency to servlet-api 4.0. I'm not sure why they did that, seems there's no apparent reason ( https://github.com/graphql-java-kickstart/graphql-java-servlet/commit/c4f680e28647e5242cf3446728640d382dd50112 )
- new dataloader 2.2.3 is required, seems to be ok
- transitive dependencies requires again updates of jackson to 2.12.x . Not a real issue, but this also require an update of snakeyaml 1.27

The main issue is the missing exports, which prevents the provider to provide anything. All in all , it will require forks to repackage the jars.

In terms of code upgrade, we need to split the provider into query/mutation/subscription ( https://github.com/Jahia/graphql-core/pull/164#discussion_r690587190 ) . This may have impact on the way the services are wired.

FieldEvaluator is still a minor issue, that could easily be fixed once the rest is starting.

gflores-jahia:
Closing for now until we have v17 support for annotation and serlvet packages

gflores-jahia:
Closing for now until we have v17 support for annotation and serlvet packages

PR #172, opened 2021-08-25
DMF-4879 : add cypresse test

jsinovassin:
https://jira.jahia.org/browse/BACKLOG-16356


dgriffon (changes requested):
I would be better to create your own content (then remove it at the end) on which you validate the boolean value instead of checking an existing one. 
Can you add a boolean value into the JCR and to the same check?

Fgerthoffert (approved):
I would also rename the filename of the test to something more meaningful

PR #190, opened 2021-10-15
Disable Maven integration tests as config is unstable

rknj:
Attempt to fix them with https://github.com/Jahia/graphql-core/pull/153


Fgerthoffert:
Do you know why they are unstable ? I believe those were pretty stable until we started to prepare for 8.1, it could be worth figuring out the root cause.

tdraier (approved):

PR #206, opened 2021-11-15
BACKLOG-16505 - Create Cypress test to replace GraphQLNodeTypesTest

rknj:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16505


tests/cypress/integration/api/jcr/nodeTypes.ts
```diff
@@ -0,0 +1,386 @@
+import gql from 'graphql-tag'
```
gflores-jahia, line 1: Might need to rename to nodeTypes.spec.ts

gflores-jahia (changes requested):

gflores-jahia (approved):

PR #215, opened 2021-12-13
Updated to the latest version of the orb

Fgerthoffert:
The plan here is to optimize storage usage.


cedmail (dismissed):

Fgerthoffert:
sorry, I'm not sure why your review got dismissed.

I'm merging since the jobs are now passing.

PR #216, opened 2021-12-15
QA-13998 : add field to get homepage for a site

dgriffon:
https://jira.jahia.org/browse/QA-13998
Add homePage field on Site query to get the home page node
example:
```
jcr {
            nodeByPath(path: $path) {
                site {
                    homePage {
                        name
                        path
                    }
                }
            }
        }
```


tdraier (dismissed):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/site/GqlJcrSite.java
```diff
@@ -135,4 +135,11 @@ public String getDefaultLanguage() {
         return result;
     }
 
+    @GraphQLField
+    @GraphQLName("homePage")
+    @GraphQLDescription("Returns the node of the home page")
+    public GqlJcrNode getHomePage() throws RepositoryException {
+        return new GqlJcrNodeImpl(siteNode.getHome());
```
tdraier, line 142: you should not instantiate GqlJcrNodeImpl drectly, you can use `SpecializedTypeHandler.getNode`

dgriffon: Thanks for the hint!

PR #221, opened 2022-01-21
BACKLOG-16477: Restore error handling behaviour in nodesById/nodesByPath

tdraier:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16477

## Description

The behaviour of nodesById/nodesByPath was broken during graphql upgrade, because of missing ExecutionContext . The correct way to send error is to create a DataFetcherResult object which embeds both data and errors. 
Handle annotations for DataFetcherResult same way as Publisher (unboxing parameterized type).
I also fixed registration of DXGraphQLProvider as a TypeFunction in defaultTypeFunction, as it was never unregistered (and instances staking up in defaultTypeFunction on every module restart). Using a dedicated TypeFunction, registered/unregistered by osgi.


cedmail (approved):

gflores-jahia (commented):
Thanks for the clean up, I did try to figure out a way how to throw multiple errors associated with the data and didn't know it was possible through `DataFetcherResult` so good to know.

Now that we're using that approach maybe we can also do the clean up now and remove `AggregateDataFetchingException` and its references including [this](https://github.com/Jahia/graphql-core/blob/master/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/JahiaDataFetchingExceptionHandler.java#L94) since we have no use for it (afaik this is the only two instances we use it).

If not we can create a follow-up ticket.

tdraier:
Ok i've removed the `AggregateDataFetchingException` class which was indeed not used anymore .

gflores-jahia (approved):

PR #225, opened 2022-02-08
QA-14154 : display logs for all exceptions

jsinovassin:
https://jira.jahia.org/browse/QA-14154


jkevan (approved):
I hope there will not be too much log regarding errors that was not displayed before.

PR #233, opened 2022-04-01
QA-14062 : add locale session parameter

jsinovassin:
https://jira.jahia.org/browse/QA-14062

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrQuery.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrQuery.java
@@ -15,32 +15,45 @@
  */
 package org.jahia.modules.graphql.provider.dxm.node;
 
-import graphql.GraphQLError;
-import graphql.annotations.annotationTypes.*;
+import graphql.annotations.annotationTypes.GraphQLDefaultValue;
+import graphql.annotations.annotationTypes.GraphQLDescription;
+import graphql.annotations.annotationTypes.GraphQLField;
+import graphql.annotations.annotationTypes.GraphQLName;
+import graphql.annotations.annotationTypes.GraphQLNonNull;
 import graphql.annotations.connection.GraphQLConnection;
 import graphql.execution.DataFetcherResult;
-import graphql.execution.ExecutionPath;
 import graphql.schema.DataFetchingEnvironment;
 import org.apache.commons.lang.LocaleUtils;
-import org.jahia.modules.graphql.provider.dxm.*;
+import org.jahia.modules.graphql.provider.dxm.BaseGqlClientException;
+import org.jahia.modules.graphql.provider.dxm.DataFetchingException;
+import org.jahia.modules.graphql.provider.dxm.GqlConstraintHandler;
+import org.jahia.modules.graphql.provider.dxm.JahiaDataFetchingExceptionHandler;
 import org.jahia.modules.graphql.provider.dxm.predicate.FieldFiltersInput;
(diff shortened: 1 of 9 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeHelper.java
```diff
@@ -156,11 +156,12 @@ private static boolean hasPropertyValue(JCRNodeWrapper node, String language, St
...
      */
     public static JCRNodeWrapper getNodeInLanguage(JCRNodeWrapper node, String language) throws RepositoryException {
-        if (language == null) {
+        Locale sessionLocale = node.getSession().getLocale();
+        if (language == null && sessionLocale == null) {
             return node;
         }
         String workspace = node.getSession().getWorkspace().getName();
-        Locale locale = LanguageCodeConverters.languageCodeToLocale(language);
+        Locale locale = language == null ? sessionLocale : LanguageCodeConverters.languageCodeToLocale(language);
```
jkevan, line 164: This looks strange. if language == null, then we use the locale of the session of the current node, so the current node is already in the current locale ? 
I dont understand why is this necessary ?

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeQueryExtensions.java
```diff
@@ -61,14 +61,16 @@ public String getValue() {
...
      */
     @GraphQLField
     @GraphQLName("jcr")
     @GraphQLNonNull
     @GraphQLDescription("JCR Queries")
-    public static GqlJcrQuery getJcr(@GraphQLName("workspace") @GraphQLDescription("The name of the workspace to fetch the node from; either EDIT, LIVE, or null to use EDIT by default") Workspace workspace) {
-        return new GqlJcrQuery(workspace != null ? workspace : Workspace.EDIT);
+    public static GqlJcrQuery getJcr(@GraphQLName("workspace") @GraphQLDescription("The name of the workspace to fetch the node from; "
+            + "either EDIT, LIVE, or null to use EDIT by default") Workspace workspace,
+            @GraphQLName("language") @GraphQLDescription("The language to use in ") String language) {
```
jkevan, line 73: Description seem's incomplete.

jkevan (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrQuery.java
```diff
@@ -281,6 +303,9 @@ private GqlJcrNode getGqlNodeById(String uuid) throws RepositoryException {
     }
 
     private JCRSessionWrapper getSession() throws RepositoryException {
+        if(this.language != null){
+            return getSession(this.language);
+        }
```
tdraier, line 308: In general it should  be ok, the only issue I see i that a node fetched with a "localized" JcrQuery will be somewhat different than a node fetched with an classical JcrQuery : i18n properties will be readable without having to set the language on the property field . This may be seen as an improvement (that can be discussed, as it will change the api behaviour just by adding a locale at /jcr level), but would potentially break the cache as these 2 nodes (i18n/non i18n) will share the same cache key. That's the only issue I can see, maybe it will introduce other problems. 
Ideally we should add the locale only for the visibility check, but still use the node without locale ...

tdraier (changes requested):

dgriffon (dismissed):

jkevan (approved):

PR #265, opened 2022-10-13
QA-14485 Add filter to handle multiple consecutive directives

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-14485

## Description

Basic filter to prevent multiple consecutive directives which can be used for attack

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/security/MultiReadRequestWrapper.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/security/MultiReadRequestWrapper.java
@@ -1,0 +1,61 @@
+package org.jahia.modules.graphql.provider.dxm.security;
+
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+import javax.servlet.ServletInputStream;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+
+import org.apache.commons.io.IOUtils;
+
+/**
+ * Wrapper class for the HttpServletRequest, that enables reading the request data input stream more than once.
+ *
+ */
+public class MultiReadRequestWrapper extends HttpServletRequestWrapper {
+    private ByteArrayOutputStream cachedBytes;
+
+    public MultiReadRequestWrapper(HttpServletRequest request) {
+        super(request);
+    }
+
+    @Override
+    public ServletInputStream getInputStream() throws IOException {
+        if (cachedBytes == null)
+            cacheInputStream();
+
+        return new CachedServletInputStream();
+    }
+
+    @Override
+    public BufferedReader getReader() throws IOException{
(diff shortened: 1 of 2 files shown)


AKarmanov:
I need to improve regex the one used is too permissive

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/security/MultiReadRequestWrapper.java
```diff
@@ -0,0 +1,61 @@
+package org.jahia.modules.graphql.provider.dxm.security;
```
bpapez, line 1: Shouldn't we add these two classes to a separate module? @fpral says in the linked ticket

> we need to find a mitigation way for the following Jahia versions:
> 
>     Jahia 7.3.9 and 7.3.10
>     Jahia 8.0.3
>     Jahia 8.1.1 and 8.1.2
> 

Putting it into graphql-dxm-provider, we would need to create minor versions for all possible or used graphql-dxm-provider version on these mentioned Jahia versions. Having the filter in an own module we could more easily give it to customers, who found and are concerned by the known DoS vulnerability or even decide to pro-actively communicate that all customers should install it.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/security/QueryFilter.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.*;
+import java.util.regex.Pattern;
+
+import static java.util.stream.Collectors.joining;
+
+@Component(service = {Filter.class}, property = {"pattern=/graphql"}, immediate = true)
+public class QueryFilter implements Filter {
+
+    // Detect strings such as @lala@lala@lala... and with space(s) in between @lala @lala
+    private Pattern regex = Pattern.compile("@[^ ]+@|@[^ ]+[ ]+@");
```
bpapez, line 16: Could there be valid queries, where @ is used more than once in a string? As the performance issue only happens with many thousands of tokens, could we perhaps use regular expression, which allows a certain number of tokens to not filter out valid queries?

bpapez (changes requested):
we need to find a mitigation way for the following Jahia versions:

    Jahia 7.3.9 and 7.3.10
    Jahia 8.0.3
    Jahia 8.1.1 and 8.1.2

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/security/MultiReadRequestWrapper.java
AKarmanov: Sure, I can try to move it out into a separate module

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/security/QueryFilter.java
AKarmanov: I did some investigating and you're right we can have multiple directives, I'll adjust regex. I think if we aim for say 10 we should be ok or do you have another number in mind?

PR #273, opened 2022-11-18
QA-14234 : use default fallback if present on the site of the node

jsinovassin:
https://jira.jahia.org/browse/QA-14234

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeHelper.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeHelper.java
@@ -154,6 +154,20 @@ public class NodeHelper {
      * @throws RepositoryException in case of JCR access errors
      */
     public static JCRNodeWrapper getNodeInLanguage(JCRNodeWrapper node, String language) throws RepositoryException {
+        return getNodeInLanguage(node, language, false);
+    }
+
+    /**
+     * Retrieves the JCR node using localized session. If <code>language</code> is <code>null</code>, returns the node itself.
+     *
+     * @param node the node to be retrieved in a localized session
+     * @param language the language of the localized session to use
+     * @param useFallbackLanguage if true and that the site of the node have a default locale, this locale will be used to get a translated node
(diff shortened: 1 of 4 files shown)


tdraier (dismissed):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeHelper.java
```diff
@@ -163,10 +177,18 @@ public static JCRNodeWrapper getNodeInLanguage(JCRNodeWrapper node, String langu
...
         }
         Locale locale = LanguageCodeConverters.languageCodeToLocale(language);
-        JCRSessionWrapper session = JCRSessionFactory.getInstance().getCurrentUserSession(workspace, locale);
+        JCRSessionWrapper session = JCRSessionFactory.getInstance().getCurrentUserSession(workspace, locale,
+                useFallbackLanguage  ? getFallbackLanguage(node): null);
         return session.getNodeByIdentifier(node.getIdentifier());
     }
 
+    private static Locale getFallbackLanguage(JCRNodeWrapper node) throws RepositoryException {
+        if (node.getResolveSite().isMixLanguagesActive()) {
```
jkevan, line 186: De we want to add safety check on `node.getResolveSite()` being null in case people ask for it on non site nodes ? 
Normally there should be a fallback on `systemSite`, but just in case something goes wrong it could return `null`

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeHelper.java
```diff
@@ -163,10 +177,18 @@ public static JCRNodeWrapper getNodeInLanguage(JCRNodeWrapper node, String langu
...
+            return LanguageCodeConverters.languageCodeToLocale(node.getResolveSite().getDefaultLanguage());
```
jkevan, line 187: Double call to `node.getResolveSite()` could be avoided by storing first call in a variable.

jkevan (changes requested):

jkevan (approved):

PR #277, opened 2022-11-24
BACKLOG-16838: Add api to fetch translation languages

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-16838

## Description

Add api to fetch languages of translations available for a node

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNode.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNode.java
@@ -271,8 +271,10 @@ public interface GqlJcrNode {
 
     @GraphQLField
     @GraphQLName("translationLanguages")
-    @GraphQLDescription("Returns nodes of available translations for this node")
-    public List<String> getTranslationLanguages();
+    @GraphQLDescription("Returns languages of available translations for this node")
+    public List<String> getTranslationLanguages(
+            @GraphQLName("isActiveOnly") @GraphQLDescription("Optional: Return languages only if it is active for the site") Boolean isActiveOnly
+    );
 
     /**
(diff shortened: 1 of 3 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNode.java
```diff
@@ -269,6 +269,11 @@ String getAggregatedLastModifiedDate(@GraphQLName("language") @GraphQLDescriptio
     List<String> getLanguagesToTranslate(@GraphQLName("languagesTranslated") @GraphQLDescription("List of known translated languages, will be used to compare modifications dates") List<String> languagesTranslated,
                                    @GraphQLName("languagesToCheck") @GraphQLDescription("List of languages potentially to be translated") List<String> languagesToCheck);
 
+    @GraphQLField
+    @GraphQLName("translationLanguages")
+    @GraphQLDescription("Returns nodes of available translations for this node")
```
cedmail, line 274: it should say return language of translated node

cedmail: We might also want to pass a boolean to be able to filter list against active languages on the site

cedmail (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNode.java
gflores-jahia: Added isActiveOnly param to filter active languages only; also added tests for both scenarios (all an active only)

cedmail (approved):

PR #324, opened 2023-05-17
BACKLOG-20591: Implement GraphQL API to get cluster DB journal state

jkevan:


jsinovassin (dismissed):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/admin/GqlClusterJournal.java
```diff
@@ -0,0 +1,116 @@
...
+            if (globalRevision == null) {
+                throw new IllegalStateException("Unable to check if cluster is sync, globalRevision not found");
+            }
+
+            List<Long> revisions = new ArrayList<>(queryAllLocalRevisions(session));
+            revisions.add(globalRevision);
+
+            return revisions.stream().distinct().count() <= 1;
+        }
+    }
```
cedmail, line 83: This is nice to know if the cluster is sync or not
But how do I find which node is out of sync if all I can get is the local revision.

I think we should be able to return journal_id/revision number to have a better view of the cluster status (there maybe still some old cluster_id registered for example (saw that before))

cedmail (changes requested):

cedmail (approved):

PR #332, opened 2023-08-01
BACKLOG-21071 Create subscription for publication job, filter by user…

AKarmanov:
## JIRA

https://jira.jahia.org/browse/BACKLOG-21071
## Description

Create subscription for publication job, filter by user key, return language


tdraier:
> I don't know if there is a better way to get translation language, I guess one way would be to update all the places where we create these jobs to actually include publication language. For now I grab first available translation node and check its language, should be reasonably accurate. What do you think?

I don't remember how it's done in gwt, but we also display the language in publication dashboard. There's probably a property somewhere, as gwt detects the language even if there's no translation node at all .

tdraier:
> I don't know if there is a better way to get translation language, I guess one way would be to update all the places where we create these jobs to actually include publication language. For now I grab first available translation node and check its language, should be reasonably accurate. What do you think?

Actually there's a "locale" property : https://github.com/Jahia/jahia-private/blob/25cab3d25c3f680f9dca33b7d006275131222a2a/core/src/main/java/org/jahia/ajax/gwt/helper/WorkflowHelper.java#L170

tdraier (approved):

PR #373, opened 2024-01-24
TECH-661: Move test graphqlRender from JUnit to Cypress

jayblanc:
## JIRA

https://jira.jahia.org/browse/TECH-661

## Description

Create a site to test if some nodes have the property isDisplayableNode or not. That property should be only some nodes like pages and not on others like list. 

## Tests

- [X] Unit Tests (Most changes _should_ have unit tests)
- [ ] Integration Tests

## Checklist

- [ ] Performance
- [ ] Migration
- [ ] Code maintainability

## Documentation

- [ ] Inline documentation


tests/provisioning-manifest-build.yml
```diff
@@ -1 +1,2 @@
 - sleep: 10
+- include: 'mvn:org.jahia.packages/jahia-additional-modules/${const:org.jahia.api.Constants.JAHIA_PROJECT_VERSION}/yaml/provisioning'
```
Fgerthoffert, line 2: Why running the provisioning for "jahia-additional-modules" ?

tests/provisioning-manifest-snapshot.yml
```diff
@@ -1,3 +1,4 @@
+- include: 'mvn:org.jahia.packages/jahia-additional-modules/${const:org.jahia.api.Constants.JAHIA_PROJECT_VERSION}/yaml/provisioning'
```
Fgerthoffert, line 1: Same question than above

Fgerthoffert (commented):
You also probably need to remove the test you did migrate over

tests/provisioning-manifest-build.yml
jayblanc: Test is creating a site using a template that is in a module

Fgerthoffert (approved):

PR #403, opened 2024-02-13
QA-15048 - Fix SDL extension subtype issues

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/QA-15048


gflores-jahia (commented):
Changes looks good to me. I'm just not sure if the failure on API description tests might have something to do with it.

PR #414, opened 2024-03-05
TECH-1634: Add new entry point for publication job subscription and deprecate old one.

cedmail:
## JIRA

https://jira.jahia.org/browse/TECH-1634

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
@@ -403,8 +403,4 @@ public class DXGraphQLProvider implements
         }).build();
         sdlSchemaService.setCodeRegistry(codeRegistry);
     }
-
-    public BundleContext getBundleContext() {
-        return bundleContext;
-    }
 }
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
@@ -403,4 +403,8 @@ public class DXGraphQLProvider implements
         }).build();
         sdlSchemaService.setCodeRegistry(codeRegistry);
     }
+
+    public BundleContext getBundleContext() {
+        return bundleContext;
+    }
 }
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/publication/PublicationJobSubscriptionExtension.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/publication/PublicationJobSubscriptionExtension.java
@@ -1,0 +1,89 @@
+package org.jahia.modules.graphql.provider.dxm.publication;
+
+import graphql.annotations.annotationTypes.GraphQLDescription;
+import graphql.annotations.annotationTypes.GraphQLField;
+import graphql.annotations.annotationTypes.GraphQLTypeExtension;
+import graphql.kickstart.servlet.osgi.GraphQLProvider;
+import io.reactivex.BackpressureStrategy;
+import io.reactivex.Flowable;
+import io.reactivex.FlowableEmitter;
+import org.apache.felix.utils.collections.MapToDictionary;
+import org.jahia.api.Constants;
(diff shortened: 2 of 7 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/publication/PublicationJobSubscriptionExtension.java
```diff
@@ -0,0 +1,89 @@
...
+        if (bundleContext == null && eventHandlerServiceRegistrationDone == null) {
+            bundleContext = ((DXGraphQLProvider) BundleUtils.getOsgiService(GraphQLProvider.class, null)).getBundleContext();
+            new PublicationJobSubscriptionExtension().registerEventHandler();
+        }
+        return Flowable.create(obs -> {
+            String name = UUID.randomUUID().toString();
+            listeners.put(name, obs);
+            logger.debug("Registered publication job listener {}", name);
+            obs.setCancellable(() -> {
+                logger.debug("Registered publication job listener {}", name);
```
jkevan, line 48: Registered to be changed for Unregistered

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/publication/PublicationJobSubscriptionExtension.java
```diff
@@ -0,0 +1,89 @@
...
+        props.put(EventConstants.EVENT_TOPIC, Constants.CLUSTER_BROADCAST_TOPIC_PREFIX + "/publication/done");
+
+        eventHandlerServiceRegistrationDone = bundleContext.registerService(EventHandler.class, this, new MapToDictionary(props));
+        props.put(EventConstants.EVENT_TOPIC, Constants.CLUSTER_BROADCAST_TOPIC_PREFIX + "/publication/start");
+        props.put(org.osgi.framework.Constants.SERVICE_PID, PublicationJobSubscriptionExtension.class.getName() + ".EventHandler.Start");
+
+        eventHandlerServiceRegistrationStart = bundleContext.registerService(EventHandler.class, this, new MapToDictionary(props));
+    }
+
+    private void unregisterEventHandler() {
```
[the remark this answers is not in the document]
cedmail: it's not called yet as we do not have a solution to plug on graphql entrypoint lifecycle

jkevan (changes requested):
It seem's we can listen on multiple topics for a single EventHandler, this could simplfy your implem by using a single registration.

Also I would suggest to separate the Subscription class from the OSGI EventHandler. It would improve readability of the implem and also allow to handle service registration more naturally:

```
@Component(property= { EventConstants.EVENT_TOPIC + "=" + Constants.CLUSTER_BROADCAST_TOPIC_PREFIX + "/publication/done",
                       EventConstants.EVENT_TOPIC + "=" + Constants.CLUSTER_BROADCAST_TOPIC_PREFIX + "/publication/start"})
public class PublicationJobEventListener implements EventHandler {

  @Override
  public void handleEvent(final Event event) {
        PublicationJobSubscriptionExtension.notifyListeners(event);    
  }
}
```

then in the PublicationJobSubscriptionExtension:
```
protected static void notifyListeners(Event event) {
        GqlPublicationEvent gqlPublicationEvent = new GqlPublicationEvent(
                event.getTopic().endsWith("/publication/done") ? GqlPublicationEvent.State.FINISHED : GqlPublicationEvent.State.STARTED,
                (List<String>) event.getProperty("siteKey"),
                (List<String>) event.getProperty("language"),
                (List<String>) event.getProperty("paths"),
                (List<String>) event.getProperty("user"));
        listeners.forEach((name, emitter) -> emitter.onNext(gqlPublicationEvent));
    }
```

This way the OSGI registration would be handle automatically.

jkevan (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/DXGraphQLProvider.java
```diff
@@ -403,4 +403,8 @@ private void transformCodeRegistryWithSDLDefaultFetcherFactory() {
         }).build();
         sdlSchemaService.setCodeRegistry(codeRegistry);
     }
+
+    public BundleContext getBundleContext() {
+        return bundleContext;
+    }
```
jkevan, line 409: Can be removed now.

jkevan (changes requested):

jkevan (approved):
Thx for the modifications !

PR #432, opened 2024-05-09
TECH-655: Migrated GQL reference tests to cypress

r3dm1ke:
## JIRA

https://jira.jahia.org/browse/tech-655

## Description

This PR migrates the existing GraphQL JCR References tests from Java to JS/Cypress. The test logic is untouched. I also disabled the "watch" option in cypress as it breaks the tests if you are using a debugger. Additionally, I updated the bundled prettier config to be consistent with eslint rules.

## Tests

- [x] Unit Tests (Most changes _should_ have unit tests)
- [x] Integration Tests

## Checklist

- [x] Performance
- [x] Migration
- [x] Code maintainability

## Documentation

- [ ] Inline documentation


tests/cypress/e2e/api/jcr/graphqlReferences.cy.ts
```diff
@@ -0,0 +1,250 @@
...
+                        nodeByPath(path: "/testList/reference1") {
+                            property(name: "j:node") {
+                                refNode {
+                                    name
+                                }
+                            }
+                        }
+                    }
+                }
+            `
```
gflores-jahia, line 86: There's a bit of repetition of this query with the rest of the test functions. Maybe appropriate to refactor this graphql as a fixture (tests/cypress/fixture) with parameters.

gflores-jahia:
Also I believe we want to attach this to [TECH-655](https://jira.jahia.org/browse/TECH-655) instead, so we should probably change the references to it.

PR #428, opened 2024-04-04
BACKLOG-22507: Save session user during workflow subscription

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/BACKLOG-22507

## Description

Fix issue where fetching active workflow task count on workflow subscription is done with incorrect session user. Currently it's using the user that initiated the workflow and not the user subscribed to the event.

Use the session user when gql subscription is initiated instead when querying for active workflow tasks.

PR #435, opened 2024-05-16
Update parent version in the tests

rknj:
## Description

See the title

I've used `mvn versions:update-parent` to automatically update to the most recent release version


gflores-jahia:
tests seems to be failing, kinda similiar to my PR a while ago where we had issues with provisioning https://github.com/Jahia/graphql-core/pull/428 but nightly seems to be ok. 

Could it be coming from the version? The core modules themselves have 8.2.0.0 parent. Maybe we need to stick to that instead, and to keep consistent?

gflores-jahia (approved):

PR #442, opened 2024-07-12
QA-15043 Return thumbnail url only if thumbnail node exists

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-15043

## Description

Return thumbnail url only if thumbnail node exists

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNode.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNode.java
@@ -315,7 +315,7 @@ public interface GqlJcrNode {
 
     @GraphQLField
     @GraphQLDescription("Get node thumbnail URL")
-    String getThumbnailUrl(@GraphQLName("name") @GraphQLDescription("Thumbnail name") String name, @GraphQLName("checkIfExists") @GraphQLDescription("Checks if requested thumbnail node exists") Boolean checkIfExists);
+    String getThumbnailUrl(@GraphQLName("name") @GraphQLDescription("Thumbnail name") String name, @GraphQLDefaultValue(GqlUtils.SupplierFalse.class) @GraphQLName("checkIfExists") @GraphQLDescription("Checks if requested thumbnail node exists") Boolean checkIfExists);
 
     /**
      * Nodes filter based on their types.


gflores-jahia:
@jkevan can we get your take on this one if safe to modify? We found thumbnail urls are still being returned even if it doesn't exist, so we're adding the check and returning null if there isn't one.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeImpl.java
```diff
@@ -551,7 +551,15 @@ public String getUrl() {
...
-        return node.getThumbnailUrl(name);
+        try {
+            if (node.hasNode(name)) {
+                return node.getThumbnailUrl(name);
+            }
+        } catch (RepositoryException e) {
+            throw new DataFetchingException(e);
+        }
+
+        return null;
```
jkevan, line 562: I am not really sure about the check. It's not because a node have a child node named: toto, that toto is actually a thumbnail.

There is functions available on JCRNodeWrapper to get the list of available thumbnails already:
`public List<String> getThumbnails() {`
That is doing additional checks to not return some node types.

Also I would be in favor of updating the description to reflect that in case thumbnail does not exists it will return null.

jkevan (changes requested):

jkevan:
Question what do you think about peoples that would have use this GraphQL field without expecting null to be returned ?
Would it be breaking for them ?
Should we consider adding an extra parameter to the field: `checkIfExists` to ensure no breakage ?

gflores-jahia:
Yes, that was my original concern is breaking change for previous users who was expecting it to return a url, even if that url is not valid. If we agree that the current behavior is broken and that it shouldn't return a url when it's not a valid url, then fixing it with this change and adding a param and default to old behaviour to not introduce breakage might be the best way.

gflores-jahia:
But I also agree that we can also look into exposing other more robust functions that has more checks e.g. getThumbnailUrls

jkevan:
> Yes, that was my original concern is breaking change for previous users who was expecting it to return a url, even if that url is not valid. If we agree that the current behavior is broken and that it shouldn't return a url when it's not a valid url, then fixing it with this change and adding a param and default to old behaviour to not introduce breakage might be the best way.

Either a new param, either a new GraphQL field dedicated to check existence of the thumbnails. Then client application could choose to call the url building or not based on this new field.

gflores-jahia:
From further discussion, we will aim to expose `node.getThumbnailUrls()` as a separate API as this looks to be doing extra checks that we need to do and avoid breakage, pending @AKarmanov checking if it does work for his purpose.

gflores-jahia (approved):

jkevan (changes requested):

jkevan (approved):

PR #443, opened 2024-07-23
QA-15228 Get render url

AKarmanov:
## JIRA

https://jira.jahia.org/browse/QA-15228 

## Description

This is somewhat experimental

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeImpl.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeImpl.java
@@ -34,6 +34,7 @@ import org.jahia.modules.graphql.provider.dxm.relay.PaginationHelper;
 import org.jahia.modules.graphql.provider.dxm.security.PermissionHelper;
 import org.jahia.modules.graphql.provider.dxm.util.ContextUtil;
 import org.jahia.modules.graphql.provider.dxm.util.GqlUtils;
+import org.jahia.modules.graphql.provider.dxm.util.ServletUtil;
 import org.jahia.services.content.*;
 import org.jahia.services.content.decorator.JCRSiteNode;
 import org.jahia.services.render.RenderContext;
@@ -51,7 +52,6 @@ import javax.jcr.nodetype.ConstraintViolationException;
 import javax.jcr.security.AccessControlException;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletRequestWrapper;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
@@ -60,7 +60,6 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 import javax.inject.Inject;
-import javax.servlet.http.HttpServletResponseWrapper;
 
 
 /**
@@ -591,7 +590,7 @@ public class GqlJcrNodeImpl implements GqlJcrNode {
             String url = getNodeURL(this.node, workspace.getValue(), LanguageCodeConverters.languageCodeToLocale(language), findDisplayable);
             HttpServletResponse httpServletResponse = ContextUtil.getHttpServletResponse(environment.getGraphQlContext());
             HttpServletRequest httpServletRequest = ContextUtil.getHttpServletRequest(environment.getGraphQlContext());
-            return urlRewriteService.rewriteOutbound(url, new RenderUrlHttpServletResponseWrapper(httpServletRequest), httpServletResponse);
+            return urlRewriteService.rewriteOutbound(url, ServletUtil.unwrapRequest(httpServletRequest), httpServletResponse);
         } catch (RepositoryException | ServletException | IOException | InvocationTargetException e) {
             throw new DataFetchingException(e);
         }
@@ -703,28 +702,5 @@ public class GqlJcrNodeImpl implements GqlJcrNode {
         url += nodeForURL.getPath() + extensionName;
 
         return url;
-    }
-
-    /**
-     * Custom request wrapper designed to get rid of unnecessary "/modules" context part, which may be injected in some cases
-     * by rewrite engine rules. On default Jahia installation this kind of context is injected into resulting url by outbound rule
-     * associated with org.jahia.services.seo.jcr.VanityUrlMapper when processing requests to /modules/graphql.
-     */
-    private class RenderUrlHttpServletResponseWrapper extends HttpServletRequestWrapper {
-
-        public RenderUrlHttpServletResponseWrapper(HttpServletRequest request) {
-            super(request);
-        }
-
-        @Override
-        public String getContextPath() {
-            String ctx = super.getContextPath();
-
-            if (ctx.endsWith("/modules")) {
-                ctx = ctx.replace("/modules", "");
-            }
-
-            return ctx;
-        }
     }
 }
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/render/RenderNodeExtensions.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/render/RenderNodeExtensions.java
@@ -23,6 +23,7 @@ import org.jahia.modules.graphql.provider.dxm.node.GqlJcrNode;
 import org.jahia.modules.graphql.provider.dxm.node.NodeHelper;
 import org.jahia.modules.graphql.provider.dxm.node.SpecializedTypesHandler;
 import org.jahia.modules.graphql.provider.dxm.util.ContextUtil;
+import org.jahia.modules.graphql.provider.dxm.util.ServletUtil;
 import org.jahia.services.SpringContextSingleton;
 import org.jahia.services.content.JCRContentUtils;
 import org.jahia.services.content.JCRNodeWrapper;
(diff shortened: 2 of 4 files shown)


jkevan:
Thanks you for the PR and the changes !

The existing logic in Jahia core is available in NodeHelper: https://github.com/Jahia/jahia-private/blob/master/core/src/main/java/org/jahia/ajax/gwt/helper/NodeHelper.java#L102-L139

The current GWT code consists of static methods used exclusively by GWT UIs. While it is part of the backend for GWT actions, this code is accessible to modules. NavigationHelper, being part of jahia-impl and a public class, could technically be used directly in GraphQL code, but its package is org.jahia.ajax.gwt.helper. Even though it's possible to use GWT backend code from GraphQL, I don't support this approach as it seems inappropriate, despite the potential for code reuse and avoiding duplication.

Your current approach seems reasonable to me.

I recall we faced similar challenges with Publication methods a few years ago. The key issue is how to avoid repeating past mistakes. Originally, this code should not have been tied to GWT, and now we're making a similar mistake by placing it in the GraphQL module.

If we (or customers) need to generate URLs for content nodes, they will either:

- Duplicate our Java code
- Use the GraphQL API on the client side

To address these design issues, we should consider:

- Option 1: Place this method in a Jahia backend utility class or service, such as JCRContentUtils for a static utility location or URLGenerator for a more service-oriented layer.
- Option 2: Plan to expose GraphQL API classes/methods to the Java backend, allowing pure Java code from modules to call APIs from the GraphQL layer.

I prefer Option 1. However, this would require GraphQL modules to depend on the latest version of Jahia to access these new methods.

For now, the best option is your current approach. I suggest adding a TODO in your code and creating a new JIRA tech ticket for future steps to clean up this code and move it to the appropriate service/utility class at a suitable time.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeImpl.java
```diff
@@ -562,6 +572,24 @@ public String getThumbnailUrl(@GraphQLName("name") @GraphQLDescription("Thumbnai
...
+                               @GraphQLName("language") @GraphQLDescription("The language content is rendered in") @GraphQLNonNull String language,
+                               @GraphQLDefaultValue(GqlUtils.SupplierFalse.class) @GraphQLName("findDisplayable") @GraphQLDescription("Finds displayable node") Boolean findDisplayable,
+                               DataFetchingEnvironment environment) {
+        try {
+            String url =  getNodeURL(this.node, workspace.getValue(), LanguageCodeConverters.languageCodeToLocale(language), findDisplayable);
+            HttpServletResponse httpServletResponse = ContextUtil.getHttpServletResponse(environment.getGraphQlContext());
+            HttpServletRequest httpServletRequest = ContextUtil.getHttpServletRequest(environment.getGraphQlContext());
+            UrlRewriteService urs = (UrlRewriteService) SpringContextSingleton.getBean("UrlRewriteService");
+            HttpServletResponse resp = new UrlRewriteWrappedResponse(httpServletResponse, httpServletRequest, urs.getEngine());
```
jkevan, line 586: Do we need to wrap the response here, I see that it was not done in the original GWT code ?
https://github.com/Jahia/jahia-private/blob/master/core/src/main/java/org/jahia/ajax/gwt/content/server/JahiaContentManagementServiceImpl.java#L1416

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeImpl.java
```diff
@@ -638,4 +666,33 @@ private void collectUsages(PropertyIterator references, Collection<GqlUsage> gql
...
+        if (template != null || nodeForURL.isNodeType("jnt:page") || nodeForURL.isNodeType("jmix:mainResource")) {
+            extensionName = ".html";
+        } else {
+            extensionName = ".content-template.html";
+        }
+
+        url += nodeForURL.getPath() + extensionName;
+
+        return url;
+    }
```
jkevan, line 697: Would add a TODO mentioning:

-  reference the original source code: [mention in comment that it's a copy/paste from Node](https://github.com/Jahia/jahia-private/blob/master/core/src/main/java/org/jahia/ajax/gwt/helper/NodeHelper.java#L102-L139)
- reference a tech ticket to move this code in a proper utility/service layer for later.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeImpl.java
```diff
@@ -562,6 +572,24 @@ public String getThumbnailUrl(@GraphQLName("name") @GraphQLDescription("Thumbnai
...
+                               @GraphQLName("language") @GraphQLDescription("The language content is rendered in") @GraphQLNonNull String language,
+                               @GraphQLDefaultValue(GqlUtils.SupplierFalse.class) @GraphQLName("findDisplayable") @GraphQLDescription("Finds displayable node") Boolean findDisplayable,
+                               DataFetchingEnvironment environment) {
+        try {
+            String url =  getNodeURL(this.node, workspace.getValue(), LanguageCodeConverters.languageCodeToLocale(language), findDisplayable);
+            HttpServletResponse httpServletResponse = ContextUtil.getHttpServletResponse(environment.getGraphQlContext());
+            HttpServletRequest httpServletRequest = ContextUtil.getHttpServletRequest(environment.getGraphQlContext());
+            UrlRewriteService urs = (UrlRewriteService) SpringContextSingleton.getBean("UrlRewriteService");
+            HttpServletResponse resp = new UrlRewriteWrappedResponse(httpServletResponse, httpServletRequest, urs.getEngine());
+            return resp.encodeURL(url).replace("/modules", "");
```
jkevan, line 587: this /modules replace looks strange. Not sure to understand why it's necessary, consider adding comments in code to explain this replace.

jkevan (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeImpl.java
```diff
@@ -562,6 +579,25 @@ public String getThumbnailUrl(@GraphQLName("name") @GraphQLDescription("Thumbnai
...
+                               @GraphQLDefaultValue(GqlUtils.SupplierFalse.class) @GraphQLName("findDisplayable") @GraphQLDescription("Finds displayable node") Boolean findDisplayable,
+                               DataFetchingEnvironment environment) {
+        try {
+            String url = getNodeURL(this.node, workspace.getValue(), LanguageCodeConverters.languageCodeToLocale(language), findDisplayable);
+            HttpServletResponse httpServletResponse = ContextUtil.getHttpServletResponse(environment.getGraphQlContext());
+            HttpServletRequest httpServletRequest = ContextUtil.getHttpServletRequest(environment.getGraphQlContext());
+            url = urlRewriteService.rewriteOutbound(url, httpServletRequest, httpServletResponse);
+            // Need to strip off /modules as it is being added after rewrite rules are applied. It is not needed for the
+            // final url.
+            return url.replace("/modules", "");
```
jkevan, line 595: The current replace code still look dangerous to me, as it is arbitrary stripping the String `/modules` chain from the url, no matter the place of this chain.

Are you 100% that the /modules that will be replace here will always be the one you want to remove ? Do we risk anything for node paths like this one:
`/sites/digitall/home/pageA/area/modulesPresentation/moduleA`

jkevan (changes requested):
See: https://github.com/Jahia/graphql-core/pull/443/files#r1691031631

Also due to the nature of this API and difficulties to predict how URL rewrites rules already in places are behaving when called from a `/modules/graphql` request, I would strongly suggest to implement tests in the current pull request.

I could be possible to implement Cypress tests that will ceate contents nodes in various places/scenarios:
- default, live workspaces
- servername on site
- vanity urls
- custom url rewrite rule

And check that the expected output URL from this new graphQL API: `getRenderURL` is correct.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeImpl.java
jkevan: Thanks for the comment, still I have the feeling that something is wrong about this replace, please take a look at: https://github.com/Jahia/graphql-core/pull/443#discussion_r1691031631

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeImpl.java
jkevan: One option, that could be interesting to try would consist on wrapping the currentRequest and override the .getPath to return `/cms` instead of `/modules`. 
The goal would be to make the current request done on `/modules/**` like if it was done on `/cms/**` transparently by manipulating/wrapping the request/response objects.
Not sure if it's easy or not, but just wanted to share the idea.

jkevan: Wrapper looks good indeed.

jkevan: Update, indeed the same issue happened again in the past:
Take a look at this GraphQL field: `RenderNodeExtensions.getRenderedContent(...)`
and the chosen solution seem's to be: https://github.com/Jahia/graphql-core/blob/master/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/render/RenderNodeExtensions.java#L144-L146

simply unwrap the original request from the OSGI proxy, to get access to the original request before it was transformed for the OSGI GraphQL servlet by our ServletHandler.

Could you please update your implem to replace the wrapper by the same solution from: `RenderNodeExtensions.getRenderedContent(...)` ?
Also it could be interesting to create an utility function somewhere to avoid code duplicate, considering the troubles we had to found back this solution, it could be interesting to put it somewhere useful with some code comments for the future.
And no need for additional tech ticket, current behavior is correct, and at any time we can have access to the original request.

jkevan (approved):
Thanks a lot for all the changes !
It's very good now

PR #447, opened 2024-08-27
SEC-104: Disable introspection queries in production

gflores-jahia:
## JIRA

https://jira.jahia.org/browse/SEC-104

## Description

- Disable introspection queries and graphql playground (both in developer tools and jahia tools) in production (still enabled in dev mode by default)
- Add config entry `graphql.introspection.enabled` to enable and disable for both dev and prod modes.

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/config/DXGraphQLConfig.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/config/DXGraphQLConfig.java
@@ -17,6 +17,7 @@ package org.jahia.modules.graphql.provider.dxm.config;
 
 import org.apache.commons.lang.StringUtils;
 import org.jahia.modules.graphql.provider.dxm.relay.PaginationHelper;
+import org.jahia.settings.SettingsBean;
 import org.osgi.service.cm.ConfigurationException;
 import org.osgi.service.cm.ManagedServiceFactory;
 import org.osgi.service.component.ComponentContext;
@@ -42,6 +43,8 @@ public class DXGraphQLConfig implements ManagedServiceFactory {
     private final static String CORS_ORIGINS = "http.cors.allow-origin";
     private final static String NODE_LIMIT = "graphql.fields.node.limit";
 
+    private final static String ENABLE_INTROSPECTION_MODE = "graphql.introspection.enabled";
+
     private Map<String, List<String>> keysByPid = new HashMap<>();
     private Map<String, String> permissions = new HashMap<>();
 
@@ -49,6 +52,7 @@ public class DXGraphQLConfig implements ManagedServiceFactory {
     private Map<String, Set<String>> corsOriginByPid = new HashMap<>();
 
     private int nodeLimit = 5000;
+    private boolean introspectionEnabled = false;
 
     private ComponentContext componentContext;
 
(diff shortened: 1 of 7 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/config/DXGraphQLConfig.java
```diff
@@ -131,6 +141,10 @@ public void deleted(String pid) {
         }
     }
 
+    public boolean isIntrospectionEnabled() {
+        return SettingsBean.getInstance().isDevelopmentMode() || introspectionEnabled;
```
Fgerthoffert, line 145: Just wondering about this statement, what would happen if `graphql.introspection.enabled = false`when running in development ? 
In such a case, introspection should instead be disabled.

gflores-jahia: ok I was under the impression we wanted it enabled for development all the time and that the flag only affects production. 

In this case, we can add a separate property specifically for dev and enable that by default.

gflores-jahia: After discussion, found a way to reuse the same property and apply different default values for dev and prod if not specified in the default configuration.

gflores-jahia:
should be ok to merge since introspection is only disabled for production by default.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/config/DXGraphQLConfig.java
```diff
@@ -71,6 +75,17 @@ public void updated(String pid, Dictionary<String, ?> properties) throws Configu
...
         Enumeration<String> keys = properties.keys();
+        boolean isDefaultConfig = properties.get("felix.fileinstall.filename") != null &&
+                properties.get("felix.fileinstall.filename").toString().endsWith("org.jahia.modules.graphql.provider-default.cfg");
+
+        /*
+         * Keep track of ENABLE_INTROSPECTION_MODE property if it's defined or not (using null) in the default configuration
+         * and only set it to a true/false value if it's defined in the configuration file
+         */
+        if (isDefaultConfig) {
+            introspectionEnabled = null;
```
Fgerthoffert, line 86: The graphql-dxm-provider configuration contains more than just introspection couldn't this logic be problematic, for example if the user modified "graphql.fields.node.limit" ?

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/config/DXGraphQLConfig.java
```diff
@@ -131,6 +149,13 @@ public void deleted(String pid) {
         }
     }
 
+    /**
+     * @return true if ENABLE_INTROSPECTION_MODE property is set to true for the default configuration, or false for any other value.
+     * It will be enabled in development mode by default, and disabled in production mode by default if the property is not defined.
+     */
+    public boolean isIntrospectionEnabled() {
+        return (introspectionEnabled == null && SettingsBean.getInstance().isDevelopmentMode()) || introspectionEnabled;
```
jayblanc, line 157: Isn't there is a NPE if not in dev mode and instrospectionEnabled == null   ??
Mayeb the expected behviour if dev mode override if : 
Return SettingsBean.getInstance().isDevelopmentMode() || (introspectionEnabled != null && introspectionEnabled)

r3dm1ke:
I have tested the cases and it works as expected. I rewrote the condition to prevent a potential NPE.

jayblanc (approved):

PR #448, opened 2024-09-04
QABACKLOG-1533: migrate GraphQLPermissionsTest from Selenium to Cypress

baptistegrimaud:
## JIRA

https://jira.jahia.org/browse/QABACKLOG-1533

## Description

Migrate the Selenium tests `GraphQLPermissionsTest` to Cypress.

## Checklist

- [x] Performance
- [x] Migration
- [x] Code maintainability

## Documentation

- [ ] Inline documentation


tests/cypress/e2e/api/graphqlPermissions.cy.ts
```diff
@@ -0,0 +1,182 @@
...
+                grantRoles('/testList', ['privileged'], 'testUser', 'USER');
+                revokeRoles('/testList/testSubList1', ['owner'], 'testUser', 'USER');
+                revokeRoles('/testList/reference1', ['owner'], 'testUser', 'USER');
+            }
+            );
+    });
+
+    after('Delete user and test data', () => {
+        deleteUser('testUser');
+        cy.apolloClient() // TODO needed but I don't know why
```
baptistegrimaud, line 72: I can't figure out why the credentials have to be reset (by default when creating an `apolloClient` with no parameters).
Without it, the temp credentials used in the rest of the tests are used (testUser/testPassword)

jayblanc: Remove the Todo from the comment. Only the explanation as a comment is excepted  ; there is nothing to do now.
// reset apollo client authentication credentials to avoid that following tests use the ones in that one.

tests/cypress/e2e/api/graphqlPermissions.cy.ts
```diff
@@ -0,0 +1,182 @@
...
+    });
+
+    it('Should get error not retrieve protected node', () => {
+        cy.apolloClient({username: 'testUser', password: 'testPassword'})
+            .apollo({
+                query: gql`
+                query {
+                    currentUser {
+                        username
+                    }
```
jayblanc, line 88: If you don't test anything on the current user maybe it can be removed from the query to avoid confusion and focus on the test. 
Anyway you can also add assertion on the username..

baptistegrimaud: you're right, adding the `currentUser` here is useless, removing it

jayblanc (approved):

PR #464, opened 2024-10-02
QA-15289: GraphQL JCRSite query with restricted home page

baptistegrimaud:
## JIRA

https://jira.jahia.org/browse/QA-15289

## Description

Gracefully handle scenarios where the home page has restricted access by returning `null` for the `homePage` field of the GraphQL JCRSite query.

Depends on https://github.com/Jahia/graphql-core/pull/465.

## Tests

- [x] Integration Tests

## Checklist

- [x] Performance
- [x] Migration
- [x] Code maintainability

## Documentation

- [x] Inline documentation
- [x] User-facing Documentation

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/site/GqlJcrSite.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/site/GqlJcrSite.java
@@ -110,9 +110,10 @@ public class GqlJcrSite extends GqlJcrNodeImpl implements GqlJcrNode {
 
     @GraphQLField
     @GraphQLName("homePage")
-    @GraphQLDescription("Returns the node of the home page")
+    @GraphQLDescription("Returns the node of the home page or null if no home page is defined or is not accessible")
     public GqlJcrNode getHomePage() throws RepositoryException {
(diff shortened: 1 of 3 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/user/GqlPrincipal.java
```diff
@@ -46,7 +47,9 @@ static DXPaginatedData<GqlGroup> getGroupMembership(String localPath,
 
         List<String> paths = groupManagerService.getMembershipByPath(localPath);
         Stream<GqlGroup> stream = paths.stream()
-                .map(path -> groupManagerService.lookupGroupByPath(path).getJahiaGroup())
+                .map(groupManagerService::lookupGroupByPath)
```
baptistegrimaud, line 50: `lookupGroupByPath()` can return `null` (if the group does not exist anymore).
In the new test (`site.cy.ts`), a new site is created which also creates a new group for it.
And because  `groupManagerService.getMembershipByPath(localPath)` gets the list from a cache (which for some reasons is not flushed when the site is deleted), it's returning a group that no longer exists.
Which was causing an existing test to fail (`user.cy.ts`)

jkevan (approved):

PR #467, opened 2024-10-11
PIPELINE-81/PIPELINE-79: Use re-usable workflow for tests and release

Fgerthoffert:
https://jira.jahia.org/browse/PIPELINE-79
https://jira.jahia.org/browse/PIPELINE-81

Also removed the rollback workflow since we decided not to rollback anymore


.github/workflows/on-release.yml
```diff
@@ -7,75 +7,7 @@ on:
     types: [prereleased]
 
 jobs:
-  on-release:
-    runs-on: self-hosted
+  release-module:
```
rknj, line 10: aren't we missing parameters like `github_slug` ?

Fgerthoffert: No, it remains optional.

I added some custom logic in the workflow itself to determine the slug:

https://github.com/Jahia/jahia-modules-action/blob/eade0d6a46aa25341cf5db048315bce67aa2a7ce/.github/workflows/release-module.yml#L93-L102

If none is provided, it will use the repository.

rknj (approved):

PR #472, opened 2024-12-05
QA-15375: Add deletePropertiesBatch method

AKarmanov:
## Description

Fix localized session resolution, add localized property mutation

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutationSupport.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutationSupport.java
@@ -123,12 +123,12 @@ public class GqlJcrMutationSupport {
      * @param properties the collection of properties to be deleted
      * @return Boolean
      */
-    public static boolean deleteProperties(JCRNodeWrapper node, Collection<GqlJcrPropertyI18nInput> properties) {
+    public static boolean deleteProperties(JCRNodeWrapper node, Collection<GqlJcrDeletedPropertyInput> properties) {
         if (properties == null) {
             return false;
         }
 
-        for (GqlJcrPropertyI18nInput prop : properties) {
+        for (GqlJcrDeletedPropertyInput prop : properties) {
             try {
                 NodeHelper.getNodeInLanguage(node, prop.getLanguage()).getProperty(prop.getName()).remove();
             } catch (RepositoryException e) {
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutationSupport.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutationSupport.java
@@ -117,6 +117,29 @@ public class GqlJcrMutationSupport {
     }
 
     /**
+     * Delete the provided properties to the specified node.
+     *
+     * @param node       The JCR node to delete properties from
+     * @param properties the collection of properties to be deleted
+     * @return Boolean
+     */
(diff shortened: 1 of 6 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrPropertyI18nInput.java
```diff
@@ -0,0 +1,51 @@
...
+import graphql.annotations.annotationTypes.GraphQLDescription;
+import graphql.annotations.annotationTypes.GraphQLField;
+import graphql.annotations.annotationTypes.GraphQLName;
+import graphql.annotations.annotationTypes.GraphQLNonNull;
+
+import java.util.List;
+
+@GraphQLName("JCRPropertyI18n")
+@GraphQLDescription("GraphQL representation of a JCR property to set")
+public class GqlJcrPropertyI18nInput {
```
gflores-jahia, line 27: Is it necessary to create a new class here to perform the op? Would the original GqlJcrPropertyInput would also work here if we just provide name and language only?

The class and the hierarchy is a bit strange to me as having a property class that doesn't carry at least a name/value. I also expect i18n as a specialized version of the original and the hierarchy would be the other way around, with the i18n class adding support for language.

If it's needed, we can maybe rename as something else (e.g. GqlJcrPropertyFetcherInput) so as it's not confused with a property input that doesn't support values or any of the other property params.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutationSupport.java
```diff
@@ -116,6 +116,29 @@ public static List<JCRPropertyWrapper> setProperties(JCRNodeWrapper node, Collec
...
+     * @return Boolean
+     */
+    public static boolean deleteProperties(JCRNodeWrapper node, Collection<GqlJcrPropertyI18nInput> properties) {
+        if (properties == null) {
+            return false;
+        }
+
+        for (GqlJcrPropertyI18nInput prop : properties) {
+            try {
+                NodeHelper.getNodeInLanguage(node, prop.getLanguage()).getProperty(prop.getName()).remove();
```
gflores-jahia, line 133: The way we're handling exceptions here might leave operation in a half-way state e.g. if one of them missing a language value, we end up with some props deleted and some aren't (even if rest could still be valid)

Maybe better option here is to handle exceptions individually and collect them as a single exception. This way we process all valid properties and throw an exception with list of the names not processed.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrPropertyI18nInput.java
r3dm1ke: I agree that this and `GqlJcrPropertyInput` could be either combined or tied together via inheritance

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutationSupport.java
gflores-jahia: ok, perfectly fine in that case.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrPropertyI18nInput.java
gflores-jahia: ok, agreed with the point of showing options that do not matter for the op, so yeah maybe a rename might be better, just to also avoid confusion that it's not a full property input object.

gflores-jahia (approved):

PR #508, opened 2025-04-07
chore: use reusable on code change

Janin-Michel-Mathias:
### Description
...

#### Source code

#### Tests
- [ ] I've provided Unit and/or Integration Tests

> [!TIP]


Fgerthoffert (approved):

Fgerthoffert (changes requested):
It's not a bad idea to use the reusable workflow here, just make sure that it does actually work, I'm not sure many projects are using it.

Issue #522, opened 2025-06-17
Retrieve mixin properties declared as "extends" Using GraphQL API


smonier:
### Environment and versions used

Jahia 8.X
graphql-dxm-provider 3.4.0

### Steps to reproduce

The context is : I'm working on a import utility that allows to map a json or csv file with an existing nodeType, I have access to the properties, but cannot list the mixin properties declared by extend.

I have a CND using mixin extends like:

```
[portalinsightmix:editorialBlock] mixin
   extends = portalinsightnt:portalInsight
   - editorialBlockTitle (string) internationalized
   - editorialBlockTeaser (string) internationalized
   - editorialBlockJsonData (string) indexed=no

  [portalinsightmix:importantDates] mixin
   extends = portalinsightnt:portalInsight
   - importantDatesTitle (string) internationalized
   - importantDatesTeaser (string) internationalized
   - importantDatesJsonData (string) indexed=no

[portalinsightnt:portalInsight] > jnt:content, portalinsightmix:components, mix:title
 - heroTeaser (string) mandatory internationalized
 - personalizedAdsTop (weakreference, choicelist[nodes='$currentSite/contents/portal//*;wemmix:personalizedList'])
 - personalizedAdsMiddle (weakreference, choicelist[nodes='$currentSite/contents/portal//*;wemmix:personalizedList'])
 - personalizedAdsBottom (weakreference, choicelist[nodes='$currentSite/contents/portal//*;wemmix:personalizedList'])
 - personalizedAdsRight (weakreference, choicelist[nodes='$currentSite/contents/portal//*;wemmix:personalizedList'])


```
I need to be able to list the mixin properties where querying GraphQL using nodeTypes : 

```
query GetContentPropertiesQuery($type: String!, $language: String!) {
        jcr {
            nodeTypes(filter: {includeTypes: [$type]}) {
                nodes {
                    name
                    properties(fieldFilter: {filters: [{fieldName: "hidden", value: "false"}]}) {
                        name
                        hidden
                        displayName(language: $language)
                        internationalized
                        mandatory
                        requiredType
                        constraints
                        multiple
                    }
                }
            }
        }
    }

```

### Current Behavior

Cannot retrieve mixin properties declared as extends

### Desired Behavior

I would like to be able to the mixin properties declared as extends using graphQL

### Label

#customer #S2E

smonier:
It's is a demand from Customer S2E

gflores-jahia:
[...]

PR #527, opened 2025-07-02
feat: Expose mixin extends has JCRNodeType

baptistegrimaud:
Closes https://github.com/Jahia/graphql-core/issues/522.
Follows https://github.com/Jahia/graphql-core/pull/526 based on [@jkevan's feedback](https://github.com/Jahia/graphql-core/issues/522#issuecomment-3027959243).
### Description
Expose the mixin extends of a given node type as node types, instead of only exposing their names.
A new field `mixinExtends` is created that can be used for instance like this:
```graphql
query {
  jcr {
    nodeTypes(filter: {includeTypes: ["jmix:categorized"]}) {
      nodes {
        mixinExtends {
          name
          properties {
            name
          }
      }
    }
  }
}
```

> [!TIP]


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/nodetype/GqlJcrNodeType.java
```diff
@@ -184,10 +185,18 @@ public List<GqlJcrNodeType> getSupertypes(@GraphQLName("fieldFilter") @GraphQLDe
     }
 
     @GraphQLField
-    @GraphQLName("mixinExtendsNames")
-    @GraphQLDescription("Returns the names of the mixin extends of this node type.")
-    public List<String> mixinExtendsNames() {
-        return nodeType.getMixinExtendNames();
+    @GraphQLName("mixinExtends")
+    @GraphQLDescription("Returns the mixin node types extended by this node type, filtered by the specified parameters.")
```
jkevan, line 189: Not sure that the wording is appropriate. It's actually returning the mixin node types extending this node type ?

jkevan (approved):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/nodetype/GqlJcrNodeType.java
baptistegrimaud: hmm, maybe I'm not understanding this _extends_ concept then...
With [this definition](https://github.com/Jahia/jahia/blob/6504c3ad858a47bab8d5fe3be9c043287bd11f30/test/core-test-framework/src/main/resources/WEB-INF/etc/repository/nodetypes/98-default-definitions.cnd#L257):
```
[jmix:internalLink] > jmix:templateMixin mixin
 extends = jnt:content
```
It means that `jmix:internalLink` extends `jnt:content`, right? In other words, that `jnt:content` is extended by the node type `jmix:internalLink`, no? 🤔

PR #532, opened 2025-07-25
refactor: use GraphQL default value for the workspace parameter

baptistegrimaud:
### Description
Use a `@GraphQLDefaultValue` to set a default value for both JCR queries (`NodeQueryExtensions`) and JCR mutations (`NodeMutationExtensions`).
Also, set the `workspace` parameter to the `Workspace` type in `NodeMutationExtensions` for consistency with `NodeQueryExtensions`.

> [!TIP]

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeQueryExtensions.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeQueryExtensions.java
@@ -32,7 +32,7 @@ public class NodeQueryExtensions {
      * JCR workspace to use for the operations.
      */
     @GraphQLDescription("JCR workspace to use for the operations")
-    public enum Workspace implements Supplier<Object> {
+    public enum Workspace {
 
         /**
          * Edit workspace
@@ -72,9 +72,12 @@ public class NodeQueryExtensions {
             return null;
         }
 
-        @Override
-        public Object get() {
-            return EDIT;
+        public static class DefaultWorkspaceSupplier implements Supplier<Object> {
+
+            @Override
+            public Workspace get() {
+                return EDIT;
+            }
         }
     }
 
@@ -90,7 +93,7 @@ public class NodeQueryExtensions {
     @GraphQLNonNull
     @GraphQLDescription("JCR Queries")
     public static GqlJcrQuery getJcr(@GraphQLName("workspace") @GraphQLDescription("The name of the workspace to fetch the node from; "
-            + "either EDIT, LIVE, or null to use EDIT by default") @GraphQLDefaultValue(value = Workspace.class) Workspace workspace) {
+            + "either EDIT, LIVE, or null to use EDIT by default") @GraphQLDefaultValue(value = Workspace.DefaultWorkspaceSupplier.class) Workspace workspace) {
         return new GqlJcrQuery(workspace);
     }
 }
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeMutationExtensions.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeMutationExtensions.java
@@ -37,7 +37,7 @@ public class NodeMutationExtensions {
     @GraphQLField
     @GraphQLName("jcr")
     @GraphQLDescription("JCR Mutation")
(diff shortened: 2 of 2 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeQueryExtensions.java
```diff
@@ -56,21 +58,29 @@ private Workspace(String workspace) {
...
             return workspace;
         }
+
+        public static class DefaultWorkspaceSupplier implements Supplier<Object> {
+
+            @Override
+            public Workspace get() {
+                return EDIT;
+            }
+        }
```
GauBen, line 68: damn java you verbose language, 7 lines just to get a default value

GauBen (approved):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeQueryExtensions.java
baptistegrimaud: it's actually to provide a class (needed by GraphQL) of a _Supplier_, not just a default value ;)

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
```diff
@@ -43,15 +43,39 @@
...
+    private final NodeQueryExtensions.Workspace workspace;
     private boolean save = true;
 
     /**
      * Initializes an instance of this class with the specified JCR workspace name.
      *
      * @param workspace the name of the JCR workspace
+     * @deprecated use {@link #GqlJcrMutation(NodeQueryExtensions.Workspace, boolean)} instead
      */
     public GqlJcrMutation(String workspace, boolean save) {
```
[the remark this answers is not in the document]
jkevan: Can you please complete the deprecate with an annotation: `@Deprecated(since = "X.X.X", forRemoval = true)`

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
```diff
@@ -43,15 +43,39 @@
...
+            return null;
+        }
+        if (NodeQueryExtensions.Workspace.EDIT.toString().equalsIgnoreCase(workspace)) {
+            return NodeQueryExtensions.Workspace.EDIT;
+        }
+        if (NodeQueryExtensions.Workspace.LIVE.toString().equalsIgnoreCase(workspace)) {
+            return NodeQueryExtensions.Workspace.LIVE;
+        }
+        return null;
+    }
```
jkevan, line 71: (Even if used by the deprecated code only)t
his code logically deserve to be a static method inside the `NodeQueryExtensions.Workspace` class itself.
aka: `NodeQueryExtensions.Workspace.fromName(String)`

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeQueryExtensions.java
jkevan, line 68: I wonder if we could directly implement Supplier<Object> at the Workspace class itself, would it be cleaner ? or would it be strange ?

```
@GraphQLDescription("JCR workspace to use for the operations")
    public enum Workspace implements Supplier<Object> {
        
        //..

        @Override
        public Object get() {
            return EDIT;
        }
    }
```

I let you decide :D

jkevan (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
baptistegrimaud: I actually kept it here to ensure it would get removed when removing the deprecated constructor... 😅 
But it might actually be useful in the future so makes sense to be in `NodeQueryExtensions.Workspace`!

jkevan (approved):

jkevan (approved):

baptistegrimaud:
The testing will be done as part of https://github.com/Jahia/graphql-core/pull/533

PR #533, opened 2025-07-25
feat: Track JCR operations

baptistegrimaud:
Follows https://github.com/Jahia/graphql-core/pull/532.
### Description
Add a custom HTTP header (`X-Jahia-Live-Operation`) to the GraphQL HTTP response, when JCR operations (queries/mutations) are performed on the live workspace.
The value of the header can be either:
- `query` for GraphQL queries like:
```gql
query {
    jcr {
        // rest of the query
    }
}
```
- or `mutation` for GraphQL mutations like:
```gql
query {
    mutation {
        // rest of the mutation
    }
}
```

#### Example

The following curl command:
```
curl 'http://localhost:8080/modules/graphql' \
  -u 'root:root1234' \
  -H 'content-type: application/json' \
  -H 'Referer: http://localhost:8080/' \
  --data-raw '{"variables":{},"query":"query { jcr(workspace: LIVE) { workspace } }"}' -i
```

would give something like:
```
HTTP/1.1 200 
Cache-Control: private, no-cache, no-store, must-revalidate, proxy-revalidate, max-age=0
X-Jahia-Live-Operation: query
Content-Type: application/json;charset=UTF-8
Content-Length: 37
Date: Mon, 28 Jul 2025 15:20:14 GMT

{"data":{"jcr":{"workspace":"LIVE"}}}
```

> [!TIP]


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/instrumentation/JcrLiveOperationsTrackerInstrumentation.java
```diff
@@ -0,0 +1,108 @@
...
+        HttpServletResponse response = ContextUtil.getHttpServletResponse(parameters.getGraphQLContext());
+        if (response != null && !response.isCommitted()) {
+            // Set the header only if the live workspace is used
+            if (jcrTrackerInstrumentationState.hasLiveQuery) {
+                log.debug("The GraphQL request contains a query on the live JCR workspace, setting the http header {} :{}", LIVE_OPERATION_HEADER, QUERY_OPERATION);
+                response.setHeader(LIVE_OPERATION_HEADER, QUERY_OPERATION);
+            } else if (jcrTrackerInstrumentationState.hasLiveMutation) {
+                log.debug("The GraphQL request contains a mutation on the live JCR workspace, setting the http header {} :{}", LIVE_OPERATION_HEADER, MUTATION_OPERATION);
+                response.setHeader(LIVE_OPERATION_HEADER, MUTATION_OPERATION);
+            }
```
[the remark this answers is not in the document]
baptistegrimaud: Well, I actually introduced those 2 boolean flags following [your comment](https://github.com/Jahia/jahia-private/issues/3697#issuecomment-3126194928), talking about batch queries, where it could have been possible to have both query and mutation at the same time 😄 
But it turned out it does not seem to be possible, and I forgot to simply the code afterwards...
My initial implementation actually stored the operation as you suggested: [`0cae9c9` (#533)](https://github.com/Jahia/graphql-core/pull/533/commits/0cae9c98d4e9716593508978b5d45cb8c1ecbc1e#diff-0bcaff222202da11ff0f9da325382df89600bc144db3de31a7e617b78ba28ec4R111)

But makes complete sense to refactor that as you're recommending, will do that!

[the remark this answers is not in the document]
GauBen: Great minds think alike!

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/instrumentation/JcrLiveOperationsTrackerInstrumentation.java
```diff
@@ -0,0 +1,103 @@
...
+                default:
+            }
+        }
+        return Instrumentation.super.beginField(parameters, state);
+    }
+
+    @Override
+    public CompletableFuture<ExecutionResult> instrumentExecutionResult(ExecutionResult executionResult, InstrumentationExecutionParameters parameters, InstrumentationState state) {
+        TrackerInstrumentationState jcrTrackerInstrumentationState = (TrackerInstrumentationState) state;
+        HttpServletResponse response = ContextUtil.getHttpServletResponse(parameters.getGraphQLContext());
```
jkevan, line 80: According to a comment found in: https://github.com/Jahia/graphql-core/blob/master/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/instrumentation/JCRInstrumentation.java#L60-L63

It seem's it would be safe to check for httpServletResponse to be not null

jkevan (changes requested):
Thank you for the implementation.

After reviewing the changes, I believe the current use of instrumentation may be excessive for this particular feature. While instrumentation is a powerful tool, in this case, the scope of the workspace value to be captured is limited and clearly identified. Our primary goal is to capture the workspace value from the `jcr` fields in both queries and mutations. Since this data is tightly coupled with the implementation, it would be more straightforward to handle this logic directly within the field implementations themselves:

- [`NodeMutationExtensions.java`](https://github.com/Jahia/graphql-core/blob/master/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeMutationExtensions.java)
- [`NodeQueryExtensions.java`](https://github.com/Jahia/graphql-core/blob/master/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/NodeQueryExtensions.java#L74)

There are a few reasons for preferring this approach:
- **Efficiency:** Directly accessing the required field/parameter in the implementation avoids unnecessary parsing and complexity introduced by instrumentation.
- **Readability & Maintainability:** Developers are more likely to look for this logic within the field implementations rather than in a separate instrumentation class. This makes the code easier to find, understand, and maintain.
- **Appropriate Use of Instrumentation:** Instrumentation is best used for:
  - Global logic that applies to all fields
  - Scenarios where you cannot modify the field implementation (e.g., when the code is outside your project)

Given the current requirements, I recommend extracting the workspace-handling logic into a generic utility that can be shared between the two `jcr` field implementations (query and mutation).

**Additional Consideration:**
It's also possible to retrieve a node's live counterpart directly, for example:

```graphql
query {
  jcr {
    nodeByPath(path: "/sites/mySite") {
      workspace
      nodeInWorkspace(workspace: LIVE) {
        workspace
      }
    }
  }
}
```

Should this use case also be tracked? With this approach, it's possible to switch workspaces at the node level within a single query.

baptistegrimaud:
> ```graphql
> nodeInWorkspace
> ```

Thanks for the review!

I went with this approach to make it more global & maintainable and to decouple the business logic (the actual GraphQL endpoint) from this tracking information.
That being said, the use case for now is indeed very basic and can be addressed without instrumentation by updating all endpoints using a workspace.
I'll update this PR accordingly, thanks!

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNode.java
```diff
@@ -254,7 +254,7 @@ DXPaginatedData<GqlJcrProperty> getReferences(@GraphQLName("fieldFilter") @Graph
     @GraphQLField
     @GraphQLName("nodeInWorkspace")
     @GraphQLDescription("GraphQL representation of this node in certain workspace")
-    GqlJcrNode getNodeInWorkspace(@GraphQLName("workspace") @GraphQLDescription("The target workspace") @GraphQLNonNull NodeQueryExtensions.Workspace workspace);
+    GqlJcrNode getNodeInWorkspace(@GraphQLName("workspace") @GraphQLDescription("The target workspace") @GraphQLNonNull NodeQueryExtensions.Workspace workspace, DataFetchingEnvironment environment);
```
baptistegrimaud, line 257: should that signature change considered a breaking change? 🤔

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/instrumentation/JcrLiveOperationsTrackerInstrumentation.java
baptistegrimaud: you're right, good catch!

PR #541, opened 2025-08-28
test: update graphql tests

x0h01:
- bump jahia-cypress version
- switch from cypress to internal logger and cy.step() (in graphqli18n.cy.ts _only_)
- additional (jar-specific) case for zip mutations test
- adjust runProvisioningScript() calls according to recent interface updates

### DIFF
--- a/tests/cypress/e2e/api/admin/config.cy.ts
+++ b/tests/cypress/e2e/api/admin/config.cy.ts
@@ -3,7 +3,7 @@ import gql from 'graphql-tag';
 
 describe('admin.configuration', () => {
     before('load graphql file and create test dataset', () => {
-        cy.runProvisioningScript({script: { fileName: 'admin/createConfig.json', type: 'application/json'}});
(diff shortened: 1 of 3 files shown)


tests/cypress/e2e/api/zipMutation.cy.ts
```diff
@@ -26,6 +26,13 @@ describe('Test zip mutation', () => {
         expectMimeTypeToBe(extension, expectedMimeType);
     });
 
+    it('when unzipping a file with an jar extension, should determine its MIME type based on extension (jar-specific case)', () => {
```
baptistegrimaud, line 29: out of curiosity, was an issue encountered with jar files in a zip?

baptistegrimaud (approved):

tests/cypress/e2e/api/zipMutation.cy.ts
x0h01: afair, there was a case with zipped jar being improperly detected; added this particular case during feature testing just to not miss it.

PR #542, opened 2025-09-19
ci: Migrated from Docker Hub to GitHub Packages

Fgerthoffert:
- Used ghcr.io for all -dev images and docker-openldap
 - Migrated to the cimg image alternative

### DIFF
--- a/.github/workflows/manual-run.yml
+++ b/.github/workflows/manual-run.yml
@@ -8,7 +8,7 @@ on:
       jahia_image:
         description: 'Jahia Image'
         required: true
-        default: 'jahia/jahia-ee-dev:8-SNAPSHOT'
+        default: 'ghcr.io/jahia/jahia-ee-dev:8-SNAPSHOT'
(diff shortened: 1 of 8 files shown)


gflores-jahia (approved):

Fgerthoffert:
> Looks like we might have some maven cache issues during build:
> 
> ```
> Error:  Tests run: 6, Failures: 4, Errors: 0, Skipped: 0, Time elapsed: 10.40 s <<< FAILURE! -- in TestSuite
> Error:  org.jahia.modules.graphql.security.GqlJcrPermissionCheckerTest.testPermissionOnMultipleFields on testPermissionOnMultipleFields(org.jahia.modules.graphql.security.GqlJcrPermissionCheckerTest) -- Time elapsed: 6.629 s <<< FAILURE!
> ```

Yes, I wonder if there could be an issue with previous cache that would be conflicting with what is coming from the new base container used to run the build.

PR #556, opened 2025-12-04
feat: Add introspectionCheckEnabled configuration property

gflores-jahia:
### Description

Refactoring and moving the `introspectionCheckEnabled` property from jahia.properties https://github.com/Jahia/jahia-private/pull/4526 to graphql OSGi config.

Introspection check is enabled if one of the gql configurations have the property enabled. 

This also means that to disable, users will have to make sure none of the graphql configurations have the property enabled.

#### Misc 

Remove `@typescript-eslint/no-explicit-any` in cypress tests by default instead of specifying on each file. (It's part of apollo return type definition)

#### Source code

#### Tests

> [!TIP]


AKarmanov (dismissed):

jkevan (changes requested):
Do we have a plan for switching this conf to true by default ?
Issue to be created targeting a specific Jahia/GraphQL version ? to not forget ?

jkevan (approved):

PR #561, opened 2025-12-18
test: Check that GraphQL workspace loads correctly

baptistegrimaud:
### Description
Following the issue we had with `node-polyfill-webpack-plugin` (see https://github.com/Jahia/graphql-core/pull/560 for the fix), adding a Cypress test to validate that GraphQL workspace loads successfully.

> [!TIP]

### DIFF
--- a/tests/cypress/e2e/tools/graphqlWorkspace.cy.ts
+++ b/tests/cypress/e2e/tools/graphqlWorkspace.cy.ts
@@ -1,6 +1,6 @@
 describe('GraphQL Workspace tests', () => {
     const GRAPHQL_WORKSPACE_URL = '/modules/graphql-dxm-provider/tools/graphql-workspace.jsp';
-    beforeEach('reset default value', () => {
+    beforeEach('prerequisites', () => {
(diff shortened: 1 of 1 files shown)


dgriffon (approved):
Ok for the test, I got a concern regarding the dependency to the tools modules that is not explicit in the graphql module neither on the test. In that case, is the test expected to fail if the tools module is not present?

baptistegrimaud:
> Ok for the test, I got a concern regarding the dependency to the tools modules that is not explicit in the graphql module neither on the test. In that case, is the test expected to fail if the tools module is not present?

As `tools` is listed as a core module, I assume that this module will always be present.
And that's probably also why GraphQL contains https://github.com/Jahia/graphql-core/blob/master/graphql-dxm-provider/src/main/resources/tools/graphql-workspace.jsp but without that explicit dependency defined.

The right way would be to define `tools` as a dependency of `graphql-core`... are we OK with that change (which becomes a bit more than just adding a test :) )?

dgriffon (approved):

PR #591, opened 2026-03-24
fix: Revert role permission check

gflores-jahia:
### Description

Revert changes related to requiring `adminRoles` permission to grantRole/revokeRole ACL APIs as part of story https://github.com/Jahia/jahia-private/issues/4685. 

More granular permission checks for ACL/ACE nodes and appropriate logging are now done on the JCR level as per https://github.com/Jahia/jahia-private/issues/4730.

#### Misc

We are also adjusting current snapshot version to 3.6.1-SN, to be released as a patch version at least for Jahia 8.2.3.1.


r3dm1ke (approved):
lgtm, but wanted to double check if no Jahia-depends version update is needed to get the fix from jahia core

gflores-jahia:
> lgtm, but wanted to double check if no Jahia-depends version update is needed to get the fix from jahia core

Good point, thanks for the reminder. I'll need to create the PRs after the merge

PR #597, opened 2026-04-01
feat!: New Tag Manager GraphQL backend 

smonier:
### Description
This PR adds the new Tag Manager GraphQL backend in `graphql-core` so the legacy webflow backend can be replaced by a Jahia UI extension in `jContent`, without any change in `jahia-core`.

Main changes:
- Added a new admin-scoped Tag Manager contract under:
  - `admin { jahia { tagManager(siteKey: String!) { ... } } }`
- Added read APIs for Tag Manager:
  - `tags(sortBy, sortOrder, limit, offset)`
  - `taggedContent(tag, limit, offset)`
- Added write APIs for Tag Manager:
  - `renameTag(tag, newName)`
  - `deleteTag(tag)`
  - `deleteTagOnNode(tag, nodeId)`
  - `renameTagOnNode(tag, newName, nodeId)`
- Added the supporting GraphQL types for managed tags and mutation results:
  - `ManagedTag`
  - `ManagedTagConnection`
  - `TagBulkMutationResult`
  - `TagNodeMutationResult`
  - `TagWorkspaceMutationResult`
  - `TagMutationError`
- Implemented the backend service in `graphql-dxm-provider` using only `org.jahia.services.tags.TaggingService`, as requested.
- Kept the existing `tag.suggest(...)` query unchanged for autocomplete/tag suggestions.

Implementation details:
- Reads are scoped by `siteKey` and use the current user `EDIT` session.
- Mutations validate `tagManager` permission on the site and update both `EDIT` and `LIVE`.
- Bulk operations reuse `TaggingService` and preserve the legacy behavior:
  - save every 100 processed nodes
  - final save at the end
  - cache invalidation for changed nodes
  - per-workspace error collection
- Added a dedicated single-node rename mutation so the UI no longer needs to write `j:tagList` directly through generic JCR GraphQL mutations.
- Fixed handling of tags containing special characters by removing unsafe SQL2 string interpolation for tag values and switching the Tag Manager backend to bound query values for tagged-content lookup and bulk rename/delete targeting.

Stability fixes included in this PR:
- Fixed GraphQL schema generation for the new Tag Manager DTOs by using the expected `@GraphQLField` getter pattern.
- Updated the provider packaging/manifest so the new Tag Manager types and runtime dependencies are correctly resolved at startup.
- Added the Java 17 test runtime adjustment needed for the existing provider test suite in this module.

Files/areas impacted:
- `graphql-dxm-provider/.../service/tags/*`
- `graphql-dxm-provider/.../admin/GqlJahiaAdminQuery.java`
- `graphql-dxm-provider/.../admin/GqlJahiaAdminMutation.java`
- `graphql-test/.../GraphQLTagManagerTest.java`
- `graphql-test/.../mod-graphql-test.xml`
- `graphql-dxm-provider/pom.xml`

#### Source code
- [x] I've shared and documented any breaking change
- [x] I've reviewed and updated the jahia-depends

#### Tests

Manual/integration coverage added in `graphql-test`:
- tag inventory query returns correct counts
- sort by name / occurrences
- pagination
- tagged-content query scoped by site
- bulk rename in EDIT and LIVE
- bulk delete in EDIT and LIVE
- delete on one node in EDIT and LIVE
- rename on one node in EDIT and LIVE
- special-character tag regression for view usages / rename / delete
- `tag.suggest(...)` regression remains unchanged


graphql-test/src/main/java/org/jahia/test/graphql/GraphQLTagManagerTest.java
jkevan, line 1: Thx for adding test coverage ! Unfortunately we do not want to introduce new tests of this type in the project, we actually have an initiative to migrate them to Cypress: https://github.com/Jahia/jahia-private/issues/3205
Cypress should be the testing framework for any new dev/PR

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/GqlTagManagerMutation.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public GqlTagManagerMutation(String siteKey) {
+        this.siteKey = siteKey;
+    }
+
+    @GraphQLField
+    @GraphQLDescription("Rename a tag on all tagged content under the site")
+    public GqlTagBulkMutationResult renameTag(@GraphQLName("tag") @GraphQLDescription("The tag to rename") @GraphQLNonNull String tag,
+                                              @GraphQLName("newName") @GraphQLDescription("The new tag name") @GraphQLNonNull String newName) {
+        return TagManagerService.getInstance().renameTag(siteKey, tag, newName);
```
jkevan: To inject a service in a GQL classes we can do (in case of injecting OSGI component/service) 

```
@Inject
@GraphQLOsgiService
private TagManagerService tagManagerService;
```

or just use a static pattern in case no need for an instance: `TagManagerService.doSomething(....)`

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/TagManagerService.java
```diff
@@ -0,0 +1,413 @@
...
+    private static final List<String> WORKSPACES = Arrays.asList(Constants.EDIT_WORKSPACE, Constants.LIVE_WORKSPACE);
+    private static final int READ_BATCH_SIZE = 500;
+    private static final TagManagerService INSTANCE = new TagManagerService();
+
+    private TagManagerService() {
+    }
+
+    public static TagManagerService getInstance() {
+        return INSTANCE;
+    }
```
jkevan: The manual singleton pattern (private constructor + static `INSTANCE` + `getInstance()`) is generally something we avoid in the Jahia ecosystem. It sits in an awkward middle ground:

- If this class needs references to other services (which seems to be the case), it should be an **OSGi component** with proper `@Reference` injection. A singleton can't participate in the OSGi service lifecycle — it can't declare dependencies, it can't be tracked, and it forces consumers to use `getInstance()` instead of standard injection.

- If this class is purely stateless utility logic with no dependencies, then **static methods** are simpler, more direct, and avoid the unnecessary ceremony of maintaining an instance that carries no state.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/TagManagerSortBy.java
jkevan, line 1: We already have Sort representation in GQL project, please do reuse existing concepts/code/utils as much as possible and avoid duplication.

See: https://github.com/Jahia/graphql-core/blob/cafa66156857cb7a89277cde4973b97db59dbb18/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/predicate/SorterHelper.java#L29-L44

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/GqlTagMutationError.java
```diff
@@ -0,0 +1,46 @@
...
+    public String getPath() {
+        return path;
+    }
+
+    @GraphQLField
+    @GraphQLDescription("The displayable page name when available")
+    public String getDisplayableName() {
+        return displayableName;
+    }
+}
```
jkevan: This `GqlTagMutationError` replicates fields (path, displayable name, etc.) that `GQLJCRNode` already provides. It's essentially a 1:1 translation of what the Webflow handler was building for JSP consumption, but carried over into GraphQL without rethinking the response design.

The issue is that Webflow and GraphQL have fundamentally different philosophies. Webflow handlers had to pre-build the entire response because the JSP needed everything upfront. In GraphQL, the client decides what it needs — that's the whole point.

When we create custom response types that duplicate existing graph representations, we lose that advantage and we create maintenance debt: what happens when we need more node info tomorrow? We extend `GqlTagMutationError` again, duplicating yet another field that `GQLJCRNode` already exposes.

A more GraphQL-native approach would look something like:

```graphql
{
  operation: delete
  tag: "toto"
  successfulNodesUpdated: [GQLJCRNode]
  failedNodesUpdated: [GQLJCRNode]
}
```

This buys us several things:
- **Optional leaf resolution** — clients that don't care about the affected nodes simply don't query those fields, making the response lighter and faster since node retrieval is skipped entirely
- **Full reuse of `GQLJCRNode`** — any client that does need node details gets the full power of the existing representation (path, uuid, displayable name, properties, children, etc.) for free, with no additional maintenance on our side
- **Extensibility by default** — new node-level capabilities added to `GQLJCRNode` are automatically available in every mutation response that references it

More broadly, the current GraphQL API across this module feels too focused on reproducing the Webflow handler responses as-is. I'd encourage taking a step back and rethinking how each operation fits into the existing GQL ecosystem — what types already exist, what can be composed, and what the client actually needs to query. That redesign effort is worth it and will result in a much cleaner, more maintainable API.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagManagerQuery.java
```diff
@@ -0,0 +1,53 @@
...
+        return TagManagerService.getInstance().getTags(siteKey, sortBy, sortOrder, limit, offset);
+    }
+
+    @GraphQLField
+    @GraphQLConnection(connectionFetcher = DXPaginatedDataConnectionFetcher.class)
+    @GraphQLDescription("List content items using a specific tag")
+    public DXPaginatedData<GqlJcrNode> taggedContent(@GraphQLName("tag") @GraphQLDescription("The tag to inspect") @GraphQLNonNull String tag,
+                                                     DataFetchingEnvironment environment) {
+        return TagManagerService.getInstance().getTaggedContent(siteKey, tag, environment);
+    }
```
jkevan, line 72: Probably would make more sense in non admin GQL. Could be move later after discussion.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/TagManagerService.java
```diff
@@ -0,0 +1,413 @@
...
+
+    public GqlManagedTagConnection getTags(String siteKey, TagManagerSortBy sortBy, TagManagerSortOrder sortOrder, Integer limit, Integer offset) {
+        validatePageArgument(limit, "limit");
+        validatePageArgument(offset, "offset");
```
jkevan: The existing pagination helper in project is normally generic and could already be reused to handle pagination. To be checked but it looks not normal to have to handle manually limit/offset and paginate stuff here

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/GqlManagedTagPageInfo.java
jkevan, line 1: To see if we can reuse existing generic pagination methods/helpers and not re-implement, see: https://github.com/Jahia/graphql-core/pull/597/changes#r3027604236

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagMutationResult.java
jkevan, line 1: To see if we can reuse existing generic pagination methods/helpers and not re-implement, see: https://github.com/Jahia/graphql-core/pull/597/changes#r3027604236

jkevan (changes requested):
## Overall review

Thanks for the work on this migration — moving from Webflow to GraphQL is a significant effort and the feature coverage is there. That said, the current implementation needs a rethink on several architectural fronts before we can merge. The inline comments go into specifics, but here's the big picture.

**What works well:** plugging into the admin GQL layer is the right call and makes sense for the nature of these operations. One method might fit better as a query rather than under admin, but that's a minor adjustment we can sort out before release — not a blocker.

**The GraphQL API needs to be designed as a GraphQL API, not a Webflow port.** This is the most important point of this review. The current approach translates the Webflow handler responses almost 1:1 into GQL types — custom error objects duplicating what `GQLJCRNode` already provides, pre-built response shapes instead of composable graph types, etc. GraphQL and Webflow have fundamentally different philosophies: in Webflow the server decides what the client gets, in GraphQL the client decides what it needs. The API should be redesigned to leverage the existing GQL ecosystem — reuse `GQLJCRNode`, compose response types from what already exists, and let optional leaf resolution do its job. This redesign effort is worth it and will pay off immediately in maintainability and flexibility.

**Service design patterns are inconsistent and sometimes off-track.** The singleton pattern (`getInstance()`) doesn't belong in an OSGi runtime — we have a service registry for that. Service injection via `@Inject @GraphQLOsgiService` is the standard in our GQL codebase. Similarly, existing infrastructure like the pagination helper is being reimplemented manually instead of reused. These aren't edge cases — they're core patterns of the Jahia ecosystem that should be followed consistently.

**Separation of concerns between service and API layers.** Business logic (tag retrieval, publication listeners, cache flushes) has ended up in the GraphQL layer because the core tag service doesn't provide everything needed. That's understandable, but it needs to be isolated — see the detailed comments on extracting a proper service class and tracking the core gaps.

**Package organization needs structure.** Everything lives in a single flat package — services, GQL types, GQL extensions, mappers. Splitting these into sub-packages that reflect responsibilities would make the codebase much more navigable.

**Testing should use Cypress, not the legacy framework.** We've been migrating away from the old test infrastructure — new features should be tested using Cypress.

I know this is a lot of feedback. The feature logic itself look solid — the work to be done is mostly about fitting it properly into the Jahia platform patterns and the GraphQL ecosystem.

smonier:
### Description
This PR introduces the new Tag Manager GraphQL backend used by the new `jcontent` Tag Manager UI, while refactoring the implementation to align with Jahia GraphQL and OSGi patterns.

Functional additions:
- Added Tag Manager GraphQL operations under:
  - `admin { jahia { tagManager(siteKey: String!) { ... } } }`
- Added Tag Manager read operations:
  - `tags(sortBy, sortOrder, ...)`
  - `taggedContent(tag, ...)`
- Added Tag Manager mutations:
  - `renameTag(tag, newName)`
  - `deleteTag(tag)`
  - `deleteTagOnNode(tag, nodeId)`
  - `renameTagOnNode(tag, newName, nodeId)`
- Kept existing `tag.suggest(...)` support unchanged functionally.

Architectural refactor included in this PR:
- Removed the manual singleton pattern from the Tag Manager backend.
- Replaced direct `getInstance()` usage with standard `@Inject @GraphQLOsgiService` service injection in GraphQL entry points.
- Split the implementation into dedicated services:
  - read-side service
  - mutation service
  - shared support/helper logic
- Reorganized the Tag Manager package layout into dedicated `graphql/` and `service/` subpackages.
- Replaced Webflow-style mutation DTOs duplicating node data with GraphQL-native mutation results based on `GqlJcrNode`.
- Removed custom connection/page-info wrappers and reused the provider’s generic pagination connection infrastructure.
- Reused existing sort infrastructure instead of maintaining a Tag Manager-specific sort-order enum.
- Updated the legacy `tag.suggest(...)` extension to use injected `TaggingService` and `DataFetchingException` instead of direct OSGi lookup and generic runtime exceptions.

Behavioral details:
- Read operations are scoped by `siteKey` and use the current user `EDIT` session.
- Mutations validate `tagManager` permission on the site node.
- Mutations update both `EDIT` and `LIVE` workspaces.
- Special-character tags are supported correctly by using bound SQL2 query values instead of direct interpolation.
- Cache invalidation and workspace save behavior preserve the expected bulk-processing semantics.

Notes:
- `taggedContent(...)` remains under the admin Tag Manager graph for now by explicit decision.
- No change was made to `jahia-core`.

### Checklist
#### Source code
- [x] I've shared and documented any breaking change
- [x] I've reviewed and updated the jahia-depends

#### Tests

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagWorkspaceMutationResult.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.List;
+
+@GraphQLName("TagWorkspaceMutationResult")
+@GraphQLDescription("The result of a tag mutation for one workspace")
+public class GqlTagWorkspaceMutationResult {
+    private final String workspace;
+    private final int processedCount;
+    private final List<GqlJcrNode> updatedNodes;
+    private final List<GqlJcrNode> failedNodes;
```
jkevan: **Memory concern: this could load a large number of nodes into RAM when an action impacts many nodes.**

A few thoughts on how to improve this:

**1. Reconsider what `updatedNodes` actually needs to return.**

I don't think there's much value in returning the full list of updated nodes — a simple count would likely be more relevant. I doubt the UI will ever need to list every updated node (name, path, etc.). The real value is on the failure side, where identifying specific nodes matters for manual recovery or debugging.

> *General guideline:* be cautious with batch operations that may load large numbers of elements into Java lists. Best practices are to (a) enforce limits, and (b) design the data model to carry only what's strictly necessary. Here we're holding full `JCRNodeWrapper` and GraphQL node instances, which are memory-heavy.

Looking at the previous implementation, this seems to have been used only for WebFlow messaging — we have an opportunity to redesign it in a more GraphQL/React-friendly way.

**2. Suggested approach — split into two interactions:**

- **Batch operation returns:** `X nodes updated`, `Y nodes failed`, plus the associated paths/UUIDs of failures (as plain `String`s, not node instances).
- **Initial UI render:** "X nodes updated, Y nodes failed" + a "Show failure details" button.
- **On button click:** trigger a separate GraphQL query using the failed paths/UUIDs to fetch the detailed node information.
- **Second UI render:** display the detailed failure information.

By splitting this into two distinct GraphQL/UI interactions, we offload the heavy node list from the original query, which only really needs lightweight summary data.

**3. Combine with limits** — e.g., return at most 10 failures by default, with pagination or a "show more" affordance for the rest.

(This is a very important matter that can make a huge difference on project development, make the API/system memory usage optimized is critical)

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    protected JCRNodeWrapper getAuthorizedSiteNode(String siteKey, JCRSessionWrapper session) throws RepositoryException {
+        JCRNodeWrapper siteNode = session.getNode("/sites/" + siteKey);
+        if (!siteNode.hasPermission("tagManager")) {
+            throw new DataFetchingException("Permission denied");
+        }
+
+        return siteNode;
+    }
```
jkevan: **This method is called 5 times, each time to check at the service layer whether the current user is authorized.**

A few observations:

**1. The return type is doing more than it needs to.** The method returns a `JCRNodeWrapper` of the site, but the only thing callers do with it is call `.getPath()` — which we already know, since we're forging it ourselves as `"/sites/" + siteKey`. So the returned node is essentially unused.

**2. More importantly, the authorization check should happen earlier — ideally in the `GqlTagManagerMutation` constructor.** Two reasons this is worth doing:

- **GraphQL allows composing multiple operations in a single request by design.** If a client issues a multi-mutation against `GqlTagManagerMutation` (e.g. `renameTagX`, `deleteTagY`, etc.), the authorization check is repeated on every operation. Performing the check once in the constructor would short-circuit all subsequent operations on a single failure.
- **Earlier checks simplify downstream code.** Individual operations can then trust that authorization has already been validated upstream, rather than each having to re-verify it.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.jcr.query.Query;
+import javax.jcr.query.QueryManager;
+import java.util.Locale;
+
+abstract class TagManagerServiceSupport {
+
+    protected static void flushNodeCaches(ModuleCacheProvider moduleCacheProvider, String path) {
+        moduleCacheProvider.invalidate(path, true);
+        moduleCacheProvider.flushRegexpDependenciesOfPath(path, true);
+    }
```
jkevan: **A lot of methods are carrying a `ModuleCacheProvider` reference as a parameter, but I don't see the benefit.**

The original caller obtains the instance via `ModuleCacheProvider.getInstance()` anyway, so threading it through the call chain doesn't add value — it's not enabling injection, substitution, or testability here.

I'd suggest removing the `ModuleCacheProvider` parameter from this chain of methods and simply resolving it locally where it's actually used:

```java
ModuleCacheProvider cacheProvider = ModuleCacheProvider.getInstance();
```

This keeps the method signatures cleaner and avoids the noise of passing around a singleton reference that callers don't really own.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
```diff
@@ -0,0 +1,75 @@
...
+
+    protected JCRSessionWrapper getCurrentUserEditSession() throws RepositoryException {
+        return JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE);
+    }
```
jkevan: **We usually avoid wrapping a single line of code in its own method.**

A few reasons it's worth inlining here:

- **No abstraction value.** The method body is a direct delegation — it doesn't encapsulate logic, hide complexity, or normalize a non-trivial call. The wrapped expression is already self-explanatory at the call site.
- **Indirection without payoff.** Readers have to jump to the method definition to confirm there's nothing interesting happening, which costs more attention than just reading `JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE)` inline.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerReadService.java
```diff
@@ -0,0 +1,153 @@
...
+        return nodes.stream()
+                .map(node -> {
+                    try {
+                        return SpecializedTypesHandler.getNode(node);
+                    } catch (RepositoryException e) {
+                        throw new DataFetchingException(e);
+                    }
+                })
+                .collect(Collectors.toList());
+    }
```
jkevan: **This method does too much post-processing in separate passes when it could be done in a single iteration.**

Looking at the flow:

1. `while` loop materializes all nodes into a `List<JCRNodeWrapper>`.
2. `nodes.sort(...)` does a second pass over the full list.
3. `stream().map(...).collect(...)` does a third pass to convert each `JCRNodeWrapper` to a `GqlJcrNode`.

That's three sequential passes over the same data, plus an intermediate list that only exists to be re-iterated. We can collapse this — and there are two reasonable shapes for the rewrite.

**Option A — single loop, sort at the end:**

```java
private List<GqlJcrNode> loadTaggedContent(JCRNodeWrapper siteNode, String tag, JCRSessionWrapper session) throws RepositoryException {
    NodeIterator nodeIterator = findTaggedNodes(siteNode.getPath(), tag, session);
    List<GqlJcrNode> result = new ArrayList<>();
    while (nodeIterator.hasNext()) {
        result.add(SpecializedTypesHandler.getNode((JCRNodeWrapper) nodeIterator.nextNode()));
    }
    result.sort(Comparator.comparing(gqlNode -> gqlNode.getNode().getPath()));
    return result;
}
```

**Option B — `TreeSet` with a path comparator, ordered on insertion:**

```java
private List<GqlJcrNode> loadTaggedContent(JCRNodeWrapper siteNode, String tag, JCRSessionWrapper session) throws RepositoryException {
    NodeIterator nodeIterator = findTaggedNodes(siteNode.getPath(), tag, session);
    Set<GqlJcrNode> sorted = new TreeSet<>(Comparator.comparing(n -> n.getNode().getPath()));
    while (nodeIterator.hasNext()) {
        sorted.add(SpecializedTypesHandler.getNode((JCRNodeWrapper) nodeIterator.nextNode()));
    }
    return new ArrayList<>(sorted);
}
```

Both are O(n log n). Option B reads more directly ("collect into a sorted set" rather than "collect, then sort"), but in practice `ArrayList` + `sort()` is often as fast or faster — TimSort is highly optimized and benefits from cache locality, while `TreeSet` pays for node allocation and pointer chasing on every insert. So the choice is mostly stylistic. One thing to be aware of with Option B: `TreeSet` uses the comparator for equality, which means two nodes with the same path would be silently deduplicated. JCR paths are unique by definition, so this is safe here, but worth flagging as a subtle property of the data structure.

**A few related notes:**

- **Exception handling:** wrapping `RepositoryException` in `DataFetchingException` is fine, but it's a bit inconsistent that `loadTaggedContent` declares `throws RepositoryException` while also converting some `RepositoryException`s to `DataFetchingException` mid-stream. Worth clarifying which exceptions should reach the GraphQL layer as data-fetching errors and which should propagate as repository errors.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
jkevan, line 1: **Why are we reinventing batch tag operations when `TaggingService` already provides them?**

`TaggingService` already exposes purpose-built batch APIs for exactly the operations this service is implementing:

```java
taggingService.renameTagUnderPath(startPath, session, selectedTag, tagNewName, callback);
taggingService.deleteTagUnderPath(startPath, session, selectedTag, callback);
```

These methods already handle:

- **Path-scoped iteration** over all nodes carrying the tag — no need to re-implement `findTaggedNodes` + `while (nodeIterator.hasNext())` here.
- **A pluggable `TagActionCallback<X>` hook** for per-node post-processing — which is precisely what `TagManagerActionCallback` is being used for here. The callback contract is already the right shape; we just need to implement it against the existing API instead of orchestrating the loop ourselves.

The current design re-implements all of this in `executeBulkMutation` / `executeBulkMutationInWorkspace`, with a custom `BulkNodeMutation` functional interface, manual node iteration, and ad-hoc error handling — duplicating logic that `TaggingService` was specifically designed to centralize.

**This is especially surprising because the original Webflow handler we're replacing was already doing it the right way:**

```java
public void deleteAllTags(RenderContext renderContext, MessageContext messageContext, String selectedTag) {
    JCRObservationManager.setAllEventListenersDisabled(Boolean.TRUE);
    try {
        for (String workspace : workspaces) {
            final JCRSessionWrapper session = getSystemSessionWorkspace(renderContext, workspace);
            Map<String, Set<String>> errors = taggingService.deleteTagUnderPath(
                renderContext.getSite().getJCRLocalPath(),
                session,
                selectedTag,
                new TagManagerActionCallback(moduleCacheProvider, session)
            );
            // ... error reporting ...
        }
    } catch (RepositoryException e) {
        // ...
    } finally {
        JCRObservationManager.setAllEventListenersDisabled(Boolean.FALSE);
    }
}
```

The Webflow version delegates cleanly to `taggingService.deleteTagUnderPath(...)` and lets the service handle iteration, batching, and per-node callbacks. The only meaningful difference between that and what we need here is the result shape — Webflow pushed errors into a `MessageContext` for UI rendering, while we need to return a `GqlTagWorkspaceMutationResult` for GraphQL. That's a callback-implementation concern, not a reason to rewrite the batch loop.

**Suggested direction:** keep the same delegation pattern as the Webflow handler, and adapt `TagManagerActionCallback` to produce a `GqlTagWorkspaceMutationResult` instead of `MessageContext` messages. The GraphQL service layer should own request orchestration, authorization, observation manager handling, and result shaping — not re-implement what `TaggingService` already does.

jkevan (changes requested):
**General feedback: missing Javadoc on new service classes, and missing Cypress test coverage.**

**1. Javadoc on the new service classes.**

I see a couple of new classes introduced as services — `TagManagerMutationService` and `TagManagerReadService` — and neither has any Javadoc on the class itself or on its public methods. This needs to be addressed before merging.

Documentation matters for two reasons now, not just one:

- **Human readability** — the traditional reason. Reviewers, future maintainers, and consumers of these services need to understand intent without reverse-engineering the implementation.
- **AI-assisted development** — increasingly important. Agentic coding tools (Copilot, Claude, etc.) rely heavily on Javadoc and inline comments to ground their understanding of a codebase. Well-documented service contracts dramatically improve the quality of AI-generated code that interacts with these classes — call sites, tests, refactors, related modules. Undocumented public APIs become a quality ceiling for everyone working with the codebase, human or otherwise.

Each public method on these services should have Javadoc covering:

- **What** the method does (one-line summary, then a longer description if non-trivial behavior is involved — e.g. multi-workspace iteration, observation manager toggling, partial-failure semantics).
- **Parameters** — `@param` for each, indicating expected values, whether `null` is permitted, format constraints (e.g. `siteKey` must be a valid site identifier), and any side effects on caller state.
- **Return value** — `@return` describing the shape and meaning, including what's returned on partial success / partial failure (relevant given the bulk-mutation result design).
- **Exceptions** — `@throws` for each declared exception, with the conditions that trigger it.

Class-level Javadoc should also describe the service's role, its threading expectations, and how it relates to `TaggingService` (i.e. that it's a GraphQL-facing layer, not a replacement for the underlying service).

Agentic AI tools are actually quite good at producing this kind of documentation as a first pass (But you will have to explain in details the context) running them over the new classes and then reviewing/correcting the output is usually faster than writing it by hand.

**2. Missing Cypress test coverage.**

There are no Cypress tests covering the new GraphQL mutations and queries introduced here. New functional surfaces should land with Cypress coverage from the start — adding tests retroactively tends to slip indefinitely. At minimum we should cover:

- Happy path for each mutation (`renameTag`, `deleteTag`, `renameTagOnNode`, `deleteTagOnNode`) across both workspaces.
- Authorization failure paths (unauthorized user, wrong site).
- Read-side queries surfaced by `TagManagerReadService`.

smonier:
---

All points from the review have been addressed. Summary below.

---

**1. Javadoc on new service classes**

Both `TagManagerMutationService` and `TagManagerReadService` now carry full Javadoc: class-level descriptions covering role, dual-workspace design, observation-suppression strategy, partial-failure semantics, authorization contract, and threading expectations. Every public method has `@param`, `@return`, and `@throws` entries documenting valid input constraints, null-permissibility, side effects, and what partial-success looks like in the result shape.

---

**2. Missing Cypress test coverage**

Added tagManager.cy.ts with 18 test cases across 6 `describe` blocks:

- **Read side**: `getTags` (with sorting, unknown site, auth denial), `getTaggedContent` (happy path, non-existent tag, auth denial)
- **Bulk mutations**: `renameTag` and `deleteTag` — happy path asserting both workspaces, blank-name rejection, auth denial
- **Single-node mutations**: `renameTagOnNode` and `deleteTagOnNode` — happy path with dual-workspace assertion, tag-absent no-op, cross-site rejection, auth denial

GraphQL fixture files are in tagManager.

---

**3. `GqlTagWorkspaceMutationResult` memory concern**

Redesigned. The unbounded `List<GqlJcrNode> updatedNodes` and `List<GqlJcrNode> failedNodes` fields are replaced with:

- `processedCount: Int` — sufficient for "X nodes updated" in the UI
- `failedCount: Int` — total failure count, always accurate regardless of truncation
- `failedPaths: [String]` — JCR paths of failed nodes, **capped at `MAX_REPORTED_FAILURES = 10`**

The heavy node objects for the success path are gone entirely. For failures, the UI gets a summary count up front and can trigger a follow-up `taggedContent` query with the returned paths to display full node detail on demand.

---

**4. Authorization check repeated on every operation**

Moved to the constructors of both `GqlTagManagerMutation` and `GqlTagManagerQuery`. The `tagManager` permission check on `/sites/{siteKey}` now runs once when the resolver object is instantiated, before any field method is called. All four mutation methods and both query methods trust that the check has already been validated upstream. The `getAuthorizedSiteNode()` helper in `TagManagerServiceSupport` has been removed.

---

**5. `ModuleCacheProvider` threading through method signatures**

Removed. `TagManagerServiceSupport.flushNodeCaches(String path)` now takes only the path and resolves `ModuleCacheProvider.getInstance()` internally. The parameter has been removed from `TagManagerActionCallback`'s constructor and all call sites.

---

**6. `getCurrentUserEditSession()` single-line wrapper**

Removed. `JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE)` is now called directly at each use site in `TagManagerMutationService` and `TagManagerReadService`.

---

**7. `loadTaggedContent` multi-pass iteration**

Collapsed from three passes (materialize into `List<JCRNodeWrapper>` → sort → stream-map to `GqlJcrNode`) to a single loop that converts each node to `GqlJcrNode` immediately, followed by one sort on the final list. The intermediate `List<JCRNodeWrapper>` is eliminated.

---

**8. Re-implementing bulk tag operations instead of using `TaggingService`'s existing batch API**

`renameTag` and `deleteTag` now delegate directly to `taggingService.renameTagUnderPath(...)` and `taggingService.deleteTagUnderPath(...)`, matching the pattern of the original Webflow handler. The custom `executeBulkMutation`, `executeBulkMutationInWorkspace` methods, and the `BulkNodeMutation` functional interface are removed. `TagManagerActionCallback` is wired directly to the `TagActionCallback<GqlTagWorkspaceMutationResult>` contract already provided by `TaggingService`.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagManagerMutation.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public GqlTagManagerMutation(String siteKey) {
+        this.siteKey = siteKey;
+        try {
+            if (!JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE)
+                    .getNode("/sites/" + siteKey).hasPermission("tagManager")) {
+                throw new DataFetchingException("Permission denied");
+            }
+        } catch (DataFetchingException e) {
+            throw e;
```
jkevan: No need to catch an Exception if it's directly rethrow.
You can remove this catch block.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagWorkspaceMutationResult.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+@GraphQLName("TagWorkspaceMutationResult")
+@GraphQLDescription("The result of a tag mutation for one workspace")
+public class GqlTagWorkspaceMutationResult {
+
+    /**
+     * Maximum number of failure paths included in a single response payload.
+     * Additional failures are counted in {@link #getFailedCount()} but not listed.
+     */
+    public static final int MAX_REPORTED_FAILURES = 10;
```
jkevan: I think this constant should belong to the service layer, not GQL layer.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerReadService.java
```diff
@@ -0,0 +1,225 @@
...
+    public DXPaginatedData<GqlJcrNode> getTaggedContent(String siteKey, String tag, DataFetchingEnvironment environment) {
+        try {
+            JCRSessionWrapper session = JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE);
+            String sitePath = "/sites/" + siteKey;
+            List<GqlJcrNode> nodes = loadTaggedContent(sitePath, tag, session);
+            return PaginationHelper.paginate(nodes, gqlNode -> PaginationHelper.encodeCursor(gqlNode.getUuid()), PaginationHelper.parseArguments(environment));
+        } catch (RepositoryException e) {
+            throw new DataFetchingException(e);
+        }
+    }
```
jkevan, line 167: this method is split in 3 different places:
- getTaggedContent here
- loadTaggedContent somewhere else in same class
- findTaggedNodes in TagManagerServiceSupport
For no real reasons, since each methods are only used for the current purpose.
Try to avoid this kind of unecessary split of code, it's difficult to read and follows the logic associated to the original method.
(more importantly when there is no other usages for all of them, and at the end it only relate to a very specific action/method)
In current state it's difficult to reason about the `getTaggedContent` logic

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagManagerQuery.java
```diff
@@ -0,0 +1,76 @@
...
+    public DXPaginatedData<GqlManagedTag> tags(@GraphQLName("sortBy") @GraphQLDescription("The field to sort by") TagManagerSortBy sortBy,
+                                               @GraphQLName("sortOrder") @GraphQLDescription("The sort order") SorterHelper.SortType sortOrder,
+                                               DataFetchingEnvironment environment) {
+        return tagManagerReadService.getTags(siteKey, sortBy, sortOrder, environment);
+    }
+
+    @GraphQLField
+    @GraphQLConnection(connectionFetcher = DXPaginatedDataConnectionFetcher.class)
+    @GraphQLDescription("List content items using a specific tag")
+    public DXPaginatedData<GqlJcrNode> taggedContent(@GraphQLName("tag") @GraphQLDescription("The tag to inspect") @GraphQLNonNull String tag,
```
jkevan, line 69: I wonder if we should instead handle our self limit/offset pagination at the method level, and stop rellying internally on PaginationHelper. (I know I have been pushing for it's usage...)

The main concern is the JCR query is resolving the entire nodes without limit/offest at query level, which cannot scale, and could put pressure on the server in case of large volumes. In scenarios with tag being used 10x, 100x, 1000x gonna be ok probably, but what will happens if a tag being used 10k, 100k ? It will probably crash or be very very slow.

So for the current Pr and works, that was about migrating an existing screen, it's probably not the perfect place to do such optimization.

So I would say for now, it's not worth than before, it's equality wrong.
So let's focus on the objectif, and keep such improvement for later.
(more importantly it concerns the entire GraphQL APIs that do pagination.)

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagManagerQuery.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @GraphQLField
+    @GraphQLConnection(connectionFetcher = DXPaginatedDataConnectionFetcher.class)
+    @GraphQLDescription("List tags used under the site")
+    public DXPaginatedData<GqlManagedTag> tags(@GraphQLName("sortBy") @GraphQLDescription("The field to sort by") TagManagerSortBy sortBy,
+                                               @GraphQLName("sortOrder") @GraphQLDescription("The sort order") SorterHelper.SortType sortOrder,
+                                               DataFetchingEnvironment environment) {
+        return tagManagerReadService.getTags(siteKey, sortBy, sortOrder, environment);
+    }
```
jkevan: I think we should also not re-invent sortBy, sortOrder here.
GQL API already contains a SorterHelper that allows to sort on GQL field of returned object (so any prop of GqlManagedTag).

There is couple example, see: GqlGroup:

```
@GraphQLField
    @GraphQLNonNull
    @GraphQLDescription("Group members")
    @GraphQLConnection(connectionFetcher = DXPaginatedDataConnectionFetcher.class)
    public DXPaginatedData<GqlPrincipal> getMembers(@GraphQLName("fieldFilter") @GraphQLDescription("Filter by graphQL fields values") FieldFiltersInput fieldFilter,
                                                    @GraphQLName("fieldSorter") @GraphQLDescription("Sort by graphQL fields values") FieldSorterInput fieldSorter,
                                                    @GraphQLName("fieldGrouping") @GraphQLDescription("Group fields according to specified criteria") FieldGroupingInput fieldGrouping,
                                                    DataFetchingEnvironment environment) {
        Stream<GqlPrincipal> stream = groupManagerService.lookupGroupByPath(group.getLocalPath()).getMembers().stream()
                .map(this::convertMember)
                .filter(Objects::nonNull)
                .filter(FilterHelper.getFieldPredicate(fieldFilter, FieldEvaluator.forConnection(environment)));

        if (fieldSorter != null) {
            stream = stream.sorted(SorterHelper.getFieldComparator(fieldSorter, FieldEvaluator.forConnection(environment)));
        }

        if (fieldGrouping != null) {
            stream = GroupingHelper.group(stream, fieldGrouping, FieldEvaluator.forConnection(environment));
        }

        PaginationHelper.Arguments arguments = PaginationHelper.parseArguments(environment);
        return PaginationHelper.paginate(stream, n -> PaginationHelper.encodeCursor(n.getName()), arguments);
    }
```

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -0,0 +1,315 @@
...
+     */
+    public GqlTagMutationResult renameTagOnNode(String siteKey, String tag, String newName, String nodeId) {
+        ensureMutationTagName(newName);
+        String sitePath = "/sites/" + siteKey;
+        try {
+            JCRSessionWrapper userSession = JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE);
+            validateNodeBelongsToSite(userSession, nodeId, sitePath);
+            Locale locale = userSession.getLocale();
+            Locale fallbackLocale = userSession.getFallbackLocale();
+            List<GqlTagWorkspaceMutationResult> workspaceResults = new ArrayList<>();
```
jkevan: Same here, no need for locales handling overhead.
double sessions used in a single action. make sure to use only one system session here.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
```diff
@@ -0,0 +1,67 @@
...
+        JCRNodeWrapper node = session.getNodeByIdentifier(nodeId);
+        String nodePath = node.getPath();
+        if (!nodePath.equals(sitePath) && !nodePath.startsWith(sitePath + "/")) {
+            throw new GqlJcrWrongInputException("Node does not belong to the requested site");
+        }
+    }
+
+    protected JCRSessionWrapper getSystemSession(String workspace, Locale locale, Locale fallbackLocale) throws RepositoryException {
+        return JCRSessionFactory.getInstance().getCurrentSystemSession(workspace, locale, fallbackLocale);
+    }
```
jkevan: One line method, said in a previous review should be avoided

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
```diff
@@ -0,0 +1,67 @@
...
+        cacheProvider.flushRegexpDependenciesOfPath(path, true);
+    }
+
+    protected void validateNodeBelongsToSite(JCRSessionWrapper session, String nodeId, String sitePath) throws RepositoryException {
+        JCRNodeWrapper node = session.getNodeByIdentifier(nodeId);
+        String nodePath = node.getPath();
+        if (!nodePath.equals(sitePath) && !nodePath.startsWith(sitePath + "/")) {
+            throw new GqlJcrWrongInputException("Node does not belong to the requested site");
+        }
+    }
```
jkevan: Move that in the service directly

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
```diff
@@ -0,0 +1,67 @@
...
+
+    protected NodeIterator findTaggedNodes(String sitePath, String tag, JCRSessionWrapper session) throws RepositoryException {
+        String query = "SELECT * FROM [jmix:tagged] AS result WHERE ISDESCENDANTNODE(result, '" +
+                JCRContentUtils.sqlEncode(sitePath) + "') AND (result.[j:tagList] = $tag)";
+        QueryManager queryManager = session.getWorkspace().getQueryManager();
+        Query jcrQuery = queryManager.createQuery(query, Query.JCR_SQL2);
+        Value tagValue = session.getValueFactory().createValue(tag);
+        jcrQuery.bindValue("tag", tagValue);
+        return jcrQuery.execute().getNodes();
+    }
```
jkevan: Same, only used once, move into usage place, and consider merge with main method since it's only used once, will help not scatter logic and made code coherent an easier to understand in the single place it's should belong to

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
```diff
@@ -0,0 +1,67 @@
...
+     * Invalidates all module caches associated with the given JCR path.
+     * The {@link ModuleCacheProvider} instance is resolved internally; callers do not own it.
+     *
+     * @param path the JCR path whose caches should be flushed; must not be {@code null}
+     */
+    protected static void flushNodeCaches(String path) {
+        ModuleCacheProvider cacheProvider = ModuleCacheProvider.getInstance();
+        cacheProvider.invalidate(path, true);
+        cacheProvider.flushRegexpDependenciesOfPath(path, true);
+    }
```
jkevan: Could be move in the service and keep the static nature, this way the current class could be entirely removed.

graphql-test/src/main/java/org/jahia/test/graphql/GraphQLTagManagerTest.java
jkevan: I think this class should be removed now that there is cypress test coverage.

graphql-test/src/main/resources/META-INF/spring/mod-graphql-test.xml
```diff
@@ -10,6 +10,7 @@
                 <value>org.jahia.test.graphql.GraphQLDescendantsTest</value>
                 <value>org.jahia.test.graphql.GraphQLCriteriaTest</value>
                 <value>org.jahia.test.graphql.GraphQLNodeMutationsTest</value>
+                <value>org.jahia.test.graphql.GraphQLTagManagerTest</value>
```
jkevan: To be removed

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/TagManagerSortBy.java
jkevan: This is still a problem, see: https://github.com/Jahia/graphql-core/pull/597/changes#r3236845000
we need to rely on existing sort mechanism and not re-invent it.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerReadService.java
```diff
@@ -0,0 +1,225 @@
...
+            comparator = Comparator.comparing(GqlManagedTag::getName, String.CASE_INSENSITIVE_ORDER)
+                    .thenComparing(GqlManagedTag::getName);
+        }
+
+        if (effectiveSortOrder == SorterHelper.SortType.DESC) {
+            comparator = comparator.reversed();
+        }
+
+        return comparator;
+    }
```
jkevan: May not be required, see: https://github.com/Jahia/graphql-core/pull/597/changes#r3236845000

graphql-dxm-provider/src/main/import/permissions.xml
```diff
@@ -6,5 +6,6 @@
     <admin jcr:primaryType="jnt:permission">
         <graphqlAdminQuery jcr:primaryType="jnt:permission"/>
         <graphqlAdminMutation jcr:primaryType="jnt:permission"/>
+        <tagManager jcr:primaryType="jnt:permission"/>
```
jkevan: tagManager permission have been set under: site-admin originally in `tags` module:
```
<site-admin jcr:primaryType="jnt:permission">
        <tagManager jcr:primaryType="jnt:permission"/>
    </site-admin>
```
We need to make sure and properly handle migration.
Not sure yet, how, but we should definitely avoid having two permission using the same name in the system at the same time.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -0,0 +1,315 @@
...
+     * @throws GqlJcrWrongInputException if {@code newName} is blank
+     */
+    public GqlTagMutationResult renameTag(String siteKey, String tag, String newName) {
+        ensureMutationTagName(newName);
+        String sitePath = "/sites/" + siteKey;
+        try {
+            JCRSessionWrapper userSession = JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE);
+            Locale locale = userSession.getLocale();
+            Locale fallbackLocale = userSession.getFallbackLocale();
+            List<GqlTagWorkspaceMutationResult> results = new ArrayList<>();
```
jkevan: SAme misconception:
- no need for locales overhead
- use a single system session here, not the current user session

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -0,0 +1,315 @@
...
+     *         identifier (bulk mode), and one {@link GqlTagWorkspaceMutationResult} per workspace
+     * @throws DataFetchingException wrapping a {@link RepositoryException} if session
+     *                               acquisition or query execution fails
+     */
+    public GqlTagMutationResult deleteTag(String siteKey, String tag) {
+        String sitePath = "/sites/" + siteKey;
+        try {
+            JCRSessionWrapper userSession = JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE);
+            Locale locale = userSession.getLocale();
+            Locale fallbackLocale = userSession.getFallbackLocale();
```
jkevan: SAme misconception:
- no need for locales overhead
- use a single system session here, not the current user session

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -0,0 +1,315 @@
...
+     *                                   acquisition fails
+     * @throws GqlJcrWrongInputException if the resolved node does not belong to the requested site
+     */
+    public GqlTagMutationResult deleteTagOnNode(String siteKey, String tag, String nodeId) {
+        String sitePath = "/sites/" + siteKey;
+        try {
+            JCRSessionWrapper userSession = JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE);
+            validateNodeBelongsToSite(userSession, nodeId, sitePath);
+            Locale locale = userSession.getLocale();
+            Locale fallbackLocale = userSession.getFallbackLocale();
```
jkevan: SAme misconception:

no need for locales overhead
use a single system session here, not the current user session

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+        for (org.jahia.services.content.JCRValueWrapper tagValue : node.getProperty(TaggingService.J_TAG_LIST).getValues()) {
+            if (tag.equals(tagValue.getString())) {
+                return true;
+            }
+        }
+
+        return false;
+    }
```
jkevan: This logic of testing the `nodeHasTag` before renaming is interesting, but it's seems it should belong to the TaggingService it self directly. (I know it's not in the scope of the current PR to modify the core service).
But I would prefer avoid adding such extra logic, that actually hack the service in place, and rather open a ticket on Jahia core to add this as an improvement to the existing TaggingService.
In other words, better removing this extra check from here, for now.
(also it doesn't seems to be part of existing webflow implem.)

graphql-dxm-provider/pom.xml
```diff
@@ -52,6 +52,9 @@
             org.jahia.modules.graphql.provider.dxm.sdl.parsing.status,
             org.jahia.modules.graphql.provider.dxm.sdl.registration,
             org.jahia.modules.graphql.provider.dxm.security,
+            org.jahia.modules.graphql.provider.dxm.service.tags,
+            org.jahia.modules.graphql.provider.dxm.service.tags.graphql,
+            org.jahia.modules.graphql.provider.dxm.service.tags.service,
```
jkevan: Not sure it's a good idea to export package those packages, we dont want peoples to use/extends those java classes for now. let's keep them OSGI private.

graphql-dxm-provider/pom.xml
```diff
@@ -395,6 +399,13 @@
                     <outputName>java-bom.cdx</outputName>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <argLine>--add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED</argLine>
+                </configuration>
+            </plugin>
```
jkevan, line 398: For what it is used for ?
Add comment to the code if used to understand what it is.
Or remove, if not necessary.

jkevan (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerReadService.java
smonier: Here's what changed:

TagManagerReadService.java — getTaggedContent is now self-contained: the JCR query building (from findTaggedNodes) and node collection/sorting (from loadTaggedContent) are inlined directly. The loadTaggedContent private method is removed. Added import javax.jcr.Value;.
TagManagerServiceSupport.java — findTaggedNodes method removed. Cleaned up the four imports that were only needed for it (JCRContentUtils, NodeIterator, Value, Query, QueryManager).

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagWorkspaceMutationResult.java
smonier: Here's what changed:

TagManagerActionCallback.java — MAX_REPORTED_FAILURES = 10 moved here as a public static final constant; the internal usage now references it directly without the class qualifier.
GqlTagWorkspaceMutationResult.java — Constant removed; added import for TagManagerActionCallback; updated the class-level Javadoc ({@value TagManagerActionCallback#MAX_REPORTED_FAILURES}) and the @GraphQLDescription string to reference the constant from its new home.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagManagerMutation.java
smonier: Removed the redundant catch (DataFetchingException e) { throw e; } block from GqlTagManagerMutation.java. The DataFetchingException now propagates naturally without the unnecessary catch.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagManagerQuery.java
smonier: Done nothing.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagManagerQuery.java
smonier: Here's a summary of what changed:

GqlTagManagerQuery.java — tags() now takes fieldSorter: FieldSorterInput instead of sortBy/sortOrder. Also fixed the redundant catch (DataFetchingException e) { throw e; } in the constructor (same pattern already fixed in GqlTagManagerMutation).
TagManagerReadService.java — getTags accepts FieldSorterInput and applies SorterHelper.getFieldComparator(fieldSorter, FieldEvaluator.forConnection(environment)) when non-null. loadManagedTags no longer sorts. getSortComparator deleted entirely.
TagManagerSortBy.java — deleted; no longer needed.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
smonier: Summary of changes:

TagManagerServiceSupport.java — getSystemSession simplified to getSystemSession(String workspace), calling getCurrentSystemSession(workspace, null, null). Removed Locale import.
TagManagerMutationService.java — All four public methods no longer acquire a user session or extract locale/fallbackLocale. renameTag and deleteTag call getSystemSession(workspace) directly in the loop. deleteTagOnNode and renameTagOnNode use getSystemSession(Constants.EDIT_WORKSPACE) for the site-membership validation, eliminating the double-session issue. Private helpers deleteTagOnNodeInWorkspace and renameTagOnNodeInWorkspace drop the locale params. Removed JCRSessionFactory and Locale imports.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
smonier: The getSystemSession wrapper is deleted from TagManagerServiceSupport.java along with the now-unused JCRSessionFactory import. TagManagerMutationService.java now calls JCRSessionFactory.getInstance().getCurrentSystemSession(workspace, null, null) directly at all four call sites.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
smonier: validateNodeBelongsToSite is now a private method in TagManagerMutationService.java — the only class that uses it. TagManagerServiceSupport.java now contains only flushNodeCaches.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
smonier: this was addressed earlier when we inlined findTaggedNodes (and loadTaggedContent) directly into getTaggedContent in TagManagerReadService.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerServiceSupport.java
smonier: TagManagerServiceSupport.java — deleted entirely.
TagManagerMutationService.java — no longer extends the support class; flushNodeCaches moved here as package-private static; added ModuleCacheProvider import.
TagManagerActionCallback.java — updated to call TagManagerMutationService.flushNodeCaches.
TagManagerReadService.java — no longer extends the support class; stale Javadoc reference updated.

graphql-test/src/main/resources/META-INF/spring/mod-graphql-test.xml
smonier: GraphQLTagManagerTest.java deleted — superseded by Cypress test coverage.
GraphQLTagManagerTest entry removed from mod-graphql-test.xml.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerReadService.java
smonier: TagManagerSortBy.java was deleted and the tags() query method was migrated to FieldSorterInput earlier in this session.

getSortComparator was deleted and TagManagerSortBy removed when we migrated tags() to use FieldSorterInput + SorterHelper.getFieldComparator

graphql-dxm-provider/src/main/import/permissions.xml
smonier: tagManager moved from admin to site-admin in permissions.xml, matching the original location in the Tags module.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
smonier: Already fixed — this exact block was removed in a previous change. All four public methods in TagManagerMutationService no longer acquire a user session or extract locale/fallbackLocale; they call JCRSessionFactory.getInstance().getCurrentSystemSession(workspace, null, null) directly.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
smonier: Already fixed — same change as above, applies to deleteTag.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
smonier: Already fixed — deleteTagOnNode and renameTagOnNode were both updated in the same pass. They now call JCRSessionFactory.getInstance().getCurrentSystemSession(Constants.EDIT_WORKSPACE, null, null) directly for the site validation, with no user session or locale involved.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
smonier: nodeHasTag — method deleted and the guard in renameTagOnNodeInWorkspace removed. taggingService.renameTag is now called directly; if the tag isn't present, the service handles it and any error surfaces through the existing catch block.
Stale getSystemSession calls — both deleteTagOnNodeInWorkspace and renameTagOnNodeInWorkspace now call JCRSessionFactory.getInstance().getCurrentSystemSession(workspace, null, null) directly (the earlier replace_all only matched the differently-indented calls in the public methods).

graphql-dxm-provider/pom.xml
smonier: Done. The three tag manager packages (service.tags, service.tags.graphql, service.tags.service) are removed from the OSGi export-package list in pom.xml, keeping them private to the bundle.

graphql-dxm-provider/pom.xml
smonier: Done. Added a comment to pom.xml explaining that the --add-opens flags are required because BeanWrapper calls setAccessible(true) on java.lang and java.lang.reflect members, which the Java module system blocks without them.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -0,0 +1,306 @@
...
+     *                belong to {@code /sites/{siteKey}} or a validation exception is thrown
+     * @return a {@link GqlTagMutationResult} carrying the tag name, the target {@code nodeId},
+     *         and one {@link GqlTagWorkspaceMutationResult} per workspace; on success the
+     *         processed count is 1; on failure failedCount is 1 and failedPaths contains the
+     *         node path (or nodeId as fallback)
+     * @throws DataFetchingException     wrapping a {@link RepositoryException} if session
+     *                                   acquisition fails
+     * @throws GqlJcrWrongInputException if the resolved node does not belong to the requested site
+     */
+    public GqlTagMutationResult deleteTagOnNode(String siteKey, String tag, String nodeId) {
```
jkevan, line 185: Similar to previous review case: JCRSessionFactory.getInstance().getCurrentSystemSession(Constants.EDIT_WORKSPACE, null, null) is called twice in current method. once in main body and one time in underlying method: `deleteTagOnNodeInWorkspace`
Splitting method is sometimes useful for reusability, but here the `deleteTagOnNodeInWorkspace` is only use here, so prefer not splitting.
Like mention previously splitting method increase risk of repeats, like for the current dual call to `getCurrentSystemSession`. (So in case splitting is done, we should also make sure to reuse as much a possible resources/object correctly by passing as arguments for example. But here the split does worth it, single usage, making overal deletion logic split in two for no real reason, forcing a dev to read 2 methods instead of a single one to understand the logic around the delete.)
It is very important to organize code in a smart way, splits of code per method, is not a trivial things to do randomly:
- no need split unitary logic that is self contains, like current for the deleteTagOnNode, having the entire logic in a single place, help a lot for code review, fix, code clarity.
- split if logic is getting complex to read and over intricate in the same main method (lot of hierarchy while/if nested blocks)
- split for reusability, if a code block is repeated a lot of times in multiple places, could make sense to extract it to a dedicated well defined and commented places. (not for a single line of code)

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -0,0 +1,306 @@
...
+     * @return a {@link GqlTagMutationResult} carrying the original tag name, the target
+     *         {@code nodeId}, and one {@link GqlTagWorkspaceMutationResult} per workspace;
+     *         processed count per workspace is 1 if the tag was present and renamed, 0 if the
+     *         tag was absent in that workspace
+     * @throws DataFetchingException     wrapping a {@link RepositoryException} if session
+     *                                   acquisition or infrastructure-level operations fail
+     * @throws GqlJcrWrongInputException if {@code newName} is blank or the node does not belong
+     *                                   to the requested site
+     */
+    public GqlTagMutationResult renameTagOnNode(String siteKey, String tag, String newName, String nodeId) {
```
jkevan, line 248: same as: https://github.com/Jahia/graphql-core/pull/597/changes#r3248029215

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -0,0 +1,306 @@
...
+            systemSession.save();
+            flushNodeCaches(node.getPath());
+            return new GqlTagWorkspaceMutationResult(workspace, 1, 0, Collections.emptyList());
+        } catch (RepositoryException e) {
+            String failedPath = node != null ? node.getPath() : nodeId;
+            return new GqlTagWorkspaceMutationResult(workspace, 0, 1, Collections.singletonList(failedPath));
+        }
+    }
+
+    static void flushNodeCaches(String path) {
```
jkevan, line 289: deserve a small java doc comment, it's used outside, and a dev may wonder why this method is static ? Nice to explain why.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -0,0 +1,306 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jahia.modules.graphql.provider.dxm.service.tags.service;
+
+import org.apache.commons.lang.StringUtils;
```
jkevan: use `import org.apache.commons.lang3.StringUtils` instead.
`org.apache.commons.lang` usage is not recommended.

graphql-dxm-provider/pom.xml
jkevan: When is this BeanWrapper calls to setAccessible(true) happening.
I cant see any usage of BeanWrapper in current PR.
I assume it's for the tests, but we dont want to run/maintaij java tests anymore, so I doubt we need this JVM arguments here.

graphql-test/src/main/resources/META-INF/spring/mod-graphql-test.xml
```diff
@@ -10,7 +10,7 @@
                 <value>org.jahia.test.graphql.GraphQLDescendantsTest</value>
                 <value>org.jahia.test.graphql.GraphQLCriteriaTest</value>
                 <value>org.jahia.test.graphql.GraphQLNodeMutationsTest</value>
-                <value>org.jahia.test.graphql.GraphQLPublicationTest</value>
+<value>org.jahia.test.graphql.GraphQLPublicationTest</value>
```
jkevan: Problem with indentation here ?

.vscode/settings.json
jkevan, line 1: I dont think this should be part of the commits files.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/graphql/GqlTagManagerQuery.java
jkevan: By returning public DXPaginatedData<GqlJcrNode>.
It probably means you can reuse jContent data tables to render nodes/contents on the React front end side, this way we reuse existing components for the tables and get federated table of content design (I didn't looked at front end code, Geoffrey is taking the lead on this.) ?

jkevan (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
smonier: deleteTagOnNodeInWorkspace and renameTagOnNodeInWorkspace are deleted — their logic lives directly in the public methods. Each method now acquires the edit session once, reuses it for both validateNodeBelongsToSite and the edit-workspace mutation, and gets a second session only for the live workspace.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
smonier: renameTagOnNode was inlined in the same commit that addressed deleteTagOnNode. Both private helpers were removed and their logic merged directly into the public methods.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
smonier: Javadoc added explaining both what the method does and why it's static — called by TagManagerActionCallback which has no service reference.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
smonier: Switched to org.apache.commons.lang3.StringUtils

graphql-test/src/main/resources/META-INF/spring/mod-graphql-test.xml
smonier: fixed

.vscode/settings.json
smonier: .vscode/settings.json removed from the repo and .vscode/ added to .gitignore to prevent it from being committed again.

jkevan (approved):
Thanks a lot for the time and effort fixing and addressing all code reviews !
GraphQL API is clean !
(I didn't reviewed the Cypress code, I will let a QA handle the review of this part.)

sdusolle (dismissed):

rknj (dismissed):

rknj (approved):

PR #605, opened 2026-04-22
fix: clean up cached permissions before rebuilding the schema

dgriffon:
Fixes https://github.com/Jahia/jahia-private/issues/4707
### Description
When removing (or updating) a permission of a field, the previous set permission were remaining in an internal cache. This fix clean up the cached permissions. 

Also it prevent the schema to be rebuild several times while extension are registered. 

Due to the `ReferencePolicy.STATIC` policy, at the end of a provider registration cycle, the `activate` method is called that build the schema. 

Cypress tests have been added to cover the fields change / upgrade on a module upgrade.  

#### Tests

> [!TIP]


jkevan (changes requested):
**Bug + Design Concern: `configPermissions` never propagates to `permissions` at runtime**

The intent behind splitting `configPermissions` from dynamic permissions is clear and I think it's the right direction — but there's a correctness issue in the current implementation worth addressing before this merges.

**The bug**

A config update correctly populates `configPermissions`, but those changes never make it into `permissions` at runtime. The only place `configPermissions` is merged into `permissions` is inside `clearPermission()`, which is only triggered on `@Deactivate` of `DXGraphQLProvider`. A live config update will silently have no effect on what `getPermissions()` actually returns.

**Suggested path forward**

Rather than patching the merge timing, I'd suggest going further with the split and making the two concerns unambiguously distinct in code:

- Rename `permissions` → `annotationPermissions` to make it explicit that this map is exclusively populated by `JahiaGraphQLFieldRetriever` via annotations
- Keep `configPermissions` owned entirely by `DXGraphQLConfig` updates
- Restrict the `@Deactivate` flush to `annotationPermissions` only — config permissions don't need flushing and shouldn't be caught in that lifecycle

Then update `getPermissions()` to merge both sources at read time:

```java
public Map<String, String> getPermissions() {
    // Config permissions added after annotation permissions to allows config permissions to override hardcoded annotations
    Map<String, String> merged = new HashMap<>(annotationPermissions);
    merged.putAll(configPermissions);
    return merged;
}
```

**Why this matters**

This approach makes the data flow easy to reason about at a glance: config permissions are always live and always returned, annotation permissions have their own isolated lifecycle with a dedicated flush. No hidden coupling between a config update and a deactivation hook.

The current implementation is subtly risky precisely because the bug is silent — config changes appear to apply but don't, which is the kind of issue that slips through testing and surfaces in production at the worst moment.

(The test code base looks overkill, did you consider other ways of registering GQL extensions than using module deployments ? OSGI Component can also be registered programmatically, not only from a module DS annotated class, a single test module could provide a button/action to register/unregister a given GQL extension at runtime, which could ease the test helpers and all the complexicity about module lifecycle handling)

dgriffon:
> **Bug + Design Concern: `configPermissions` never propagates to `permissions` at runtime**
> 
> The intent behind splitting `configPermissions` from dynamic permissions is clear and I think it's the right direction — but there's a correctness issue in the current implementation worth addressing before this merges.
> 
> **The bug**
> 
> A config update correctly populates `configPermissions`, but those changes never make it into `permissions` at runtime. The only place `configPermissions` is merged into `permissions` is inside `clearPermission()`, which is only triggered on `@Deactivate` of `DXGraphQLProvider`. A live config update will silently have no effect on what `getPermissions()` actually returns.
> 
> **Suggested path forward**
> 
> Rather than patching the merge timing, I'd suggest going further with the split and making the two concerns unambiguously distinct in code:
> 
> * Rename `permissions` → `annotationPermissions` to make it explicit that this map is exclusively populated by `JahiaGraphQLFieldRetriever` via annotations
> * Keep `configPermissions` owned entirely by `DXGraphQLConfig` updates
> * Restrict the `@Deactivate` flush to `annotationPermissions` only — config permissions don't need flushing and shouldn't be caught in that lifecycle
> 
> Then update `getPermissions()` to merge both sources at read time:
> 
> ```java
> public Map<String, String> getPermissions() {
>     // Config permissions added after annotation permissions to allows config permissions to override hardcoded annotations
>     Map<String, String> merged = new HashMap<>(annotationPermissions);
>     merged.putAll(configPermissions);
>     return merged;
> }
> ```
> 
> **Why this matters**
> 
> This approach makes the data flow easy to reason about at a glance: config permissions are always live and always returned, annotation permissions have their own isolated lifecycle with a dedicated flush. No hidden coupling between a config update and a deactivation hook.
> 
> The current implementation is subtly risky precisely because the bug is silent — config changes appear to apply but don't, which is the kind of issue that slips through testing and surfaces in production at the worst moment.
> 
> (The test code base looks overkill, did you consider other ways of registering GQL extensions than using module deployments ? OSGI Component can also be registered programmatically, not only from a module DS annotated class, a single test module could provide a button/action to register/unregister a given GQL extension at runtime, which could ease the test helpers and all the complexicity about module lifecycle handling)

Thank for the feedback, 

The design has been set intentionally to not merge the map on each call to `getPermissions()` as the permissions are read for each field of a query. 

Moving to a thread safe map will improve the implementation.

About the bug, I miss the refresh of the configuration that would have update the exposed configuration, I will update it accordingly. 

Related to the test, using help declared services within the same module will ease the set up, thank you for the hint

tests/cypress/fixtures/moduleManagement.ts
```diff
@@ -0,0 +1,188 @@
...
+            }).then((result: any) => {
+                const status = Array.isArray(result) ? result[0] : String(result);
+                return status === '.installed';
+            }),
+        {
+            timeout: timeoutMs,
+            interval: 2000,
+            errorMsg: `Bundle ${bundleKey}/${bundleVersion} did not reach '${expectedState}' within ${timeoutMs}ms`
+        }
+    );
```
jkevan: Are those still required ?

tests/jahia-module/schema-update-test-module/index.html
jkevan, line 1: Is this file required ?

.gitignore
```diff
@@ -12,3 +12,4 @@ buildNumber.properties
 /.project
 /.settings/
 .DS_Store
+tests/cypress/fixtures/modules/*.jar
```
jkevan: is it still required ?

tests/cypress/fixtures/groovy/checkModuleState.groovy
jkevan, line 1: Is it still required ?

tests/cypress/fixtures/groovy/startModule.groovy
jkevan, line 1: Cannot find any ref to this fixture in current PR, is still required ?

tests/cypress/fixtures/groovy/stopModule.groovy
jkevan, line 1: Cannot find any ref to this fixture in current PR, is still required ?

jkevan (changes requested):

dgriffon:
Removed remaining files / core from the previous test implementation

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/config/DXGraphQLConfig.java
```diff
@@ -19,42 +19,47 @@
 import org.jahia.modules.graphql.provider.dxm.relay.PaginationHelper;
 import org.osgi.service.cm.ConfigurationException;
 import org.osgi.service.cm.ManagedServiceFactory;
-import org.osgi.service.component.ComponentContext;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ConfigurationPolicy;
```
jkevan: Does not seems used in current class ?

jkevan (dismissed):

jkevan (approved):

PR #626, opened 2026-07-02
chore: Deprecate createVersion, keep it as an inert no-op

baptistegrimaud:
Closes https://github.com/Jahia/content-versioning/issues/96.

### Description

The `createVersion` field on JCR node mutations is deprecated and becomes an **inert no-op**: it always returns `false` and never creates a JCR version. Version management moves to the [content-versioning](https://github.com/Jahia/content-versioning) module (Jahia 8.2.4.0+).

It is kept in the schema rather than removed, and returns `false` rather than raising, because **this provider discards every write of a mutation that produced an error** — `JahiaMutationExecutionStrategy#completeField` calls `GqlJcrMutation.completeField()`, where `session.save()` lives, only when `executionContext.getErrors().isEmpty()`. Raising from this field would therefore drop the sibling fields of any caller still selecting it. jContent's `uploadFile` and `updateFileContent` both do, so file upload and replace would have failed outright. `false` is also what this field already returned when versioning was unavailable, so no caller regresses.

### Behaviour change

`jnt:file` is `mix:versionable` by node type (`[jnt:file] > … jmix:basemetadata …`, and `[jmix:basemetadata] > jmix:unversionedBasemetadata, mix:versionable`), so files stay versionable and **publication keeps versioning them** — `jahia.publication.versionedTypes` ships with `jnt:file`, so `needVersion()` stays true and the `published_at_*` checkpoints are unaffected.

What goes away is the extra `checkout` + `checkpoint` and the `uploaded_at_<timestamp>` label created at **upload** time. In practice that means an upload followed by a replace, with no publication in between, no longer leaves a JCR snapshot of the first revision. No shipped UI surfaces those labels (jContent's version tab is contributed by content-versioning; the GWT `getVersions`/`restoreVersion` RPCs have no client caller and the `version` grid column is not declared in any shipped column config), so the impact is limited to WebDAV version-history clients, the `<jcr:nodeVersion>` taglib, and custom code using `JCRVersionService`.

### Additional changes

- New Cypress case `tests/cypress/e2e/api/jcr/createVersion.cy.ts` — asserts `createVersion` returns `false` **without raising**, and that the enclosing mutation still persists its other writes. That second assertion is the regression guard for the save-gate behaviour above.
- The `jcrLiveOperationsTrackerInstrumentation` fixtures and spec now select `uuid` instead of `createVersion`; those tests are about the tracker response headers, not versioning, so decoupling them from a deprecated field is a cleanup.
- `schema.graphql` refreshed to reflect the latest state of the GraphQL endpoints with a fresh install of Jahia 8.2.4.0-SN.

### Follow-up

jContent drops its two now-pointless selections in https://github.com/Jahia/jcontent/pull/2641 (draft until this ships).

> [!TIP]


delallioui (dismissed):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeMutation.java
```diff
@@ -558,32 +554,10 @@ public boolean revokeRoles(
...
-                session.getWorkspace().getVersionManager().checkpoint(jcrNode.getPath());
-                versionService.addVersionLabel(jcrNode, label);
-                return true;
-            }
-        } catch (RepositoryException e) {
-            throw new DataFetchingException(e);
-        }
-
-        return false;
+        throw new DataFetchingException("Version creation is not supported. Use the content-versioning module for version management.");
```
eallais: Throwing from this field doesn't just fail the field — it discards every write in the enclosing mutation, and jContent still selects it.

- This provider deliberately saves nothing when a mutation produced any error: `JahiaMutationExecutionStrategy.completeField` calls `GqlJcrMutation.completeField()` — where `session.save()` lives — only `if (... && executionContext.getErrors().isEmpty())`. It's not incidental, it's the contract: `GraphQLNodeMutationsTest.executionStrategyNoSaveOnError` pins it by asserting that a successful `rename` in the same `jcr { … }` block as a failing one is *not* persisted.
- jContent's `uploadFile` and `updateFileContent` both select `createVersion` on `main` today. After this change, uploading or replacing a file in jContent runs the whole mutation, then drops the node, the title, the binary and the mime type on the floor — and Apollo's default `errorPolicy: 'none'` rejects the promise, so `.then(data => ({uuid: ...}))` never runs and the user sees a failed upload.
- That also undercuts the "customer adoption of this endpoint is minimal" premise in the description: the biggest consumer isn't a customer, it's a first-party module shipped with the product.
- Green CI isn't evidence against this — after this PR no test in the suite selects `createVersion` anymore, so nothing exercises the path.

Two ways out, and I'd take the first:

- Make the field an inert no-op (`return false`) and let `@GraphQLDeprecate` carry the message. That's exactly what the old code did when versioning wasn't supported, so no caller regresses, and the deprecation stays visible in the schema and in introspection.
- Or keep the throw, but land the jContent PR that drops the two selections first, and gate this on a release train that ships both together.

Either way, worth a case that runs `jcr { addNode(...) { createVersion uuid } }` and then asserts the node actually exists — that one is red on this branch as it stands.

Sources: [`JahiaMutationExecutionStrategy.java:61-62`](https://github.com/Jahia/graphql-core/blob/5a015fc6f701c583dee56f74a8737e0e18f2629d/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/JahiaMutationExecutionStrategy.java#L61-L62) (the save gate, introduced by `40b33776` "we only do save if there were no errors on execution") · `graphql-test/src/main/java/org/jahia/test/graphql/GraphQLNodeMutationsTest.java:1478-1496` on this branch · [`Jahia/jcontent` UploadItem.gql-mutations.js#L17](https://github.com/Jahia/jcontent/blob/40793a8c4fb86b3925e45a8a5b264e9ca809bf0e/src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.gql-mutations.js#L17) and [#L34](https://github.com/Jahia/jcontent/blob/40793a8c4fb86b3925e45a8a5b264e9ca809bf0e/src/javascript/JContent/ContentRoute/ContentLayout/Upload/UploadItem/UploadItem.gql-mutations.js#L34) · [Apollo Client — GraphQL error policies](https://www.apollographql.com/docs/react/data/error-handling#graphql-error-policies)

```suggestion
        return false;
```

eallais (changes requested):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrNodeMutation.java
baptistegrimaud: Thank you for catching this — I missed it when making the change, and the consequence for jContent's file upload would have been very unpleasant.

Addressed in [58789d2](https://github.com/Jahia/graphql-core/commit/58789d236eb09b4a51acefc51d65480e6399c7da): the field is now an inert no-op returning `false`, with `@GraphQLDeprecate` carrying the message, as you suggested. I also added the test case you asked for — `tests/cypress/e2e/api/jcr/createVersion.cy.ts` runs `jcr { addNode(...) { createVersion uuid } }` and asserts both that the field returns `false` without raising and that the node is actually persisted afterwards.

One note on where that test lives: I put it in the Cypress suite rather than next to `executionStrategyNoSaveOnError`, because the `graphql-test` module is built by CI but never deployed or executed — no workflow overrides `tests_path` (default `tests/`), and neither provisioning manifest installs the module. So the test that pins the save-on-error contract you cited isn't actually running today. I opened #652 for that.

I've also rewritten the PR title and description: the change is no longer breaking, and the "customer adoption is minimal" premise was wrong as you pointed out.

baptistegrimaud: Correction to one point in my reply above, now that I have the backend log from the run.

I said `graphql-test` is "built by CI but never deployed or executed". The first half of that was wrong: it **is** installed in PR runs (via the build-artifacts path, not the provisioning manifest), but it fails to resolve — `missing requirement osgi.wiring.package=org.jahia.test.bin` — so it never starts. Same outcome, wrong reason; I've corrected #652 with the log evidence.

Separately, the CI failure on this PR turned out to be branch staleness rather than anything in the change. The branch was 35 commits behind `main`, and the integration-tests job runs the **branch's** `tests/` directory against a provider built from the **merge** commit. `main` has since added an authorization guard on `backgroundJobSubscription` and updated `graphqlScheduler.cy.ts` to open the WS socket lazily so the handshake carries the session cookie; the branch still had the old eager `new SubscriptionClient(wsUrl)`, which now gets denied. Merged `main` in — that spec is current again.

The new `createVersion` case passed in both runs.

eallais (approved):

PR #630, opened 2026-07-06
fix(sdl): remove dev-only SDL source-watcher coupling to external-provider-modules (#629)

jkevan:
Fixes #629

## What

Removes the dev-only SDL **source-watcher** and its mandatory OSGi coupling from
`graphql-dxm-provider` to `external-provider-modules`.

- Deletes `SDLFileSourceMonitor` (the `ModulesSourceMonitor` implementation).
- Drops the `ModulesSourceMonitor` service registration and the `Jahia-Source-Folders` /
  on-disk-sources logic in `SDLRegistrationImpl`.
- Removes the mandatory `Import-Package: org.jahia.modules.external.modules.osgi` and the
  `external-provider-modules` Maven dependency.

## Why

That watcher only ever did anything when module **sources are mounted on disk** (Studio / local
dev) — never in production — yet it forced a **mandatory package import** on
`external-provider-modules`. That single wire placed `graphql-dxm-provider` inside the OSGi refresh
closure of the External Provider stack, so a refresh of `external-provider-vfs` cascaded (via
`external-provider-modules` → `graphql-dxm-provider`) into a **full GraphQL schema teardown and a
platform-wide module restart** (~38 modules), and could break the admin console mid-operation.
Full analysis and evidence in #629.

## What is preserved

SDL extensions are still discovered and (un)registered from each bundle's
`META-INF/graphql-extension.sdl` on bundle start/stop, and the schema is still rebuilt on change —
exactly as before. Only the source-file hot-reload (and its coupling) is removed.

## Verification

- `mvn clean package` on `graphql-dxm-provider` succeeds.
- Generated `MANIFEST.MF` no longer contains `org.jahia.modules.external.modules.osgi` in
  `Import-Package` — `graphql-dxm-provider` is no longer wired into the External Provider refresh
  closure.
- No remaining references to `ModulesSourceMonitor` / `external-provider-modules` in the module
  (source or pom); no tests referenced the removed feature.


baptistegrimaud (approved):

gflores-jahia (approved):
Looks good to me, but I had a question about versioning in the related PR in pack https://github.com/Jahia/jahia-pack-private/pull/1025#pullrequestreview-4645114244.

Don't know if we need to bump version in this PR as well.

PR #604, opened 2026-04-21
test: Add support for both Jahia release and snapshot versions in aclEntries test

gflores-jahia:
### Description

Adjustments done based on recent snapshot changes https://github.com/Jahia/graphql-core/pull/602

But need to add condition to also make it work for Nightly release runs.

PR #635, opened 2026-07-08
test: remove conditional statement (functionality was backported)

x0h01:


jkevan (dismissed):

gflores-jahia (commented):
compare-versions was also added in the deps because of this condition. since it's being removed maybe ok to also cleanup https://github.com/Jahia/graphql-core/pull/604/changes#diff-c7273ca77f452520adba2dfb738cd146b83110e7faa7b861318e89b2b56b5d56R14. wdyt ?

tests/package.json
```diff
@@ -11,7 +11,6 @@
     "@typescript-eslint/eslint-plugin": "^7.18.0",
     "@typescript-eslint/parser": "^7.18.0",
     "babel-plugin-istanbul": "^6.1.1",
-    "compare-versions": "^6.1.1",
```
jkevan, line 14: We may need to call yarn install to regenerate the yarn.lock

x0h01: I did, but lock file remained without changes.
I can see those 4 `compare-version` occurrences in lock file appeared in there even before @gflores-jahia's commit, so I decided to leave them as is

jkevan (approved):

gflores-jahia (approved):

PR #643, opened 2026-07-28
fix: treat SDL finder arguments as literal values in the generated JCR-SQL2 query

eallais:
## Summary

The SDL finder fetchers build their JCR-SQL2 statement by formatting the caller's argument into a
single-quoted literal. The three of them did not handle that formatting the same way, so an argument
was not always compared as the text it actually contains.

This aligns all three on `JCRContentUtils.sqlEncode` — the encoder this module already uses for
every other SQL2 literal it builds (`JahiaAclServiceImpl`, `TagManagerReadService`,
`VanityUrlJCRSiteExtensions`).

## Change

| Fetcher | Arguments now encoded |
|---|---|
| `StringFinderDataFetcher` | `contains`, `equals` |
| `WeakreferenceFinderDataFetcher` | `contains`, `equals` |
| `DateRangeDataFetcher` | `after`, `before` (via `castDate`) |

Two follow-on effects worth naming:

- **An empty argument works.** `sqlEncode` is empty-safe, so `equals: ""` is a usable query.
- **An explicitly null argument is treated as absent.** `containsKey` is true for a GraphQL argument
  passed as `null`, so the value is read before branching: `contains: null` falls through to the
  existing *"must have either 'contains' or 'equals'"* error, or to the `equals` branch when one was
  given.

Values without an apostrophe are formatted exactly as before.

Left as-is, deliberately: the `property` argument of `WeakreferenceFinderDataFetcher` is a GraphQL
enum built from the declared schema, so it is already a closed vocabulary and nothing changes for it.

## Verification

`graphql-extension-example` gains `myNewsByTitle`, the String counterpart of the `myNewsByDate`
finder it already declares, so both fetcher paths are reachable from the test schema.

New spec `tests/cypress/e2e/sdl/sdlFinderLiterals.cy.ts` (+ its Groovy fixture), five cases on three
`jnt:news` nodes — one plain title, one containing an apostrophe, one dated outside the queried
range. Each case that expects an empty result first queries the plain value and requires the marker
back, so a fixture that produced nothing fails the case rather than passing it.

1. `equals` on a title containing an apostrophe returns exactly that node.
2. `equals` on a value carrying quote characters returns none of the three markers.
3. `contains` on a value carrying quote characters does not return the plain node.
4. The date finder's baseline returns the two in-range nodes and not the out-of-range one.
5. The date finder given a bound carrying quote characters still does not return the out-of-range
   node.

Cases 2, 3 and 5 assert on the node set rather than on an error, because a value the query engine
cannot parse is a legitimate outcome — what matters is that the selection was not widened.

`mvn -pl graphql-dxm-provider compile` passes; `yarn lint` on the test module is clean. The spec runs
in the integration job and all five cases pass — `sdl/sdlFinderLiterals.cy.ts`, 5/5, in
[run 30442323682](https://github.com/Jahia/graphql-core/actions/runs/30442323682). The environment
failure that had been holding that job back (same signature as #632) has cleared.


tests/cypress/fixtures/groovy/prepareSDLFinderLiteralTest.groovy
```diff
@@ -0,0 +1,33 @@
+import org.jahia.services.content.*
+
+import javax.jcr.RepositoryException
+
+// Content for the SDL finder literal-handling test. Titles carry the "sdlLit" marker so the
+// assertions can single these nodes out of whatever else the instance holds.
+JCRTemplate.getInstance().doExecuteWithSystemSession(new JCRCallback() {
```
eallais, line 11: This fixture writes `jcr:title` and `desc`, and on `jnt:news` both are declared `i18n`. A system session opened without a locale is the wrong place to write them from, so the nodes this suite queries for may never be findable.

- `jnt:news` declares `- jcr:title (string) i18n mandatory` and `- desc (string, richtext) i18n` ([Jahia/news `definitions.cnd`](https://github.com/Jahia/news/blob/main/src/main/resources/META-INF/definitions.cnd)). An i18n value lives on the node's `jnt:translation` child, not on the node itself.
- The finder queries `n.[jcr:title]` on `[jnt:news]`. Jahia only rewrites that into a translation-node join when the session carries a locale — `QueryModifierAndOptimizerVisitor` gates the whole rewrite on `session.getLocale() != null`. The read side is fine here, since `FinderBaseDataFetcher.getLocale` falls back to `SettingsBean.getDefaultLocale()`. It is the write side that has no locale.
- `prepareValidityTest.groovy` in this same directory already has the pattern: it does the non-i18n properties in the plain system session, then reopens `doExecuteWithSystemSessionAsUser(root, "default", Locale.ENGLISH, …)` specifically to set `jcr:title`. Worth following it here.
- Note the existing `prepareSDLQueriesTest.groovy` is not evidence to the contrary — it sets `jcr:title` the same way, but no assertion in `sdlQueries.cy.ts` ever reads a title back, so nothing there would notice.

**The visible effect is that cases 1 and 4 go red; the quieter one is that cases 2, 3 and 5 stay green regardless — see the other comment on the spec.**

**Source:** [Jahia/news `definitions.cnd`](https://github.com/Jahia/news/blob/main/src/main/resources/META-INF/definitions.cnd) · `jahia-private core/src/main/java/org/jahia/services/query/QueryModifierAndOptimizerVisitor.java:656-658` (the `session.getLocale() != null` gate) · `graphql-dxm-provider/.../sdl/fetchers/FinderBaseDataFetcher.java:57-61` (read-side locale fallback) · `tests/cypress/fixtures/groovy/prepareValidityTest.groovy:21-26` (the repo's own pattern).

tests/cypress/e2e/sdl/sdlFinderLiterals.cy.ts
```diff
@@ -0,0 +1,123 @@
...
+    it('matches a title containing an apostrophe', () => {
+        cy.apollo({
+            errorPolicy: 'all',
+            query: byTitle('equals: "sdlLit O\'Brien marker"')
+        }).then(response => {
+            expect(titlesOf(response, 'myNewsByTitle')).to.deep.equal([QUOTED]);
+        });
+    });
+
+    it('keeps an equals query to its literal match when the value carries quote characters', () => {
```
eallais, line 84: These three negative cases only ever assert absence, so they are green whenever the query returns nothing — including when it returns nothing because the fixture failed, the content was not published, or the property is not where the finder looks.

- That is not hypothetical: the i18n problem in the fixture produces exactly that state, and these cases would sail through it while cases 1 and 4 fail. A reader seeing three of five green would reasonably conclude the encoding is covered.
- `titlesOf` returns `[]` on an errored response too, by design, which is right for tolerating an unparseable bound — but it means "widened" and "nothing happened at all" are indistinguishable to the assertion.
- Case 4 already does the right thing for the date path: it pins the in-range nodes with `include` before asserting the out-of-range one is absent. The equals and contains cases need the same spine.

**Worth asserting the baseline inside each negative case — query the literal value first, confirm it returns the marker, then fire the quote-carrying value and confirm it does not widen. That way a broken fixture fails the case that is supposed to catch it.**

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/sdl/fetchers/StringFinderDataFetcher.java
```diff
@@ -96,11 +96,11 @@ public Stream<GqlJcrNode> getStream(DataFetchingEnvironment environment) {
                 throw new DataFetchingException(String.format("Entry point %s must have either 'contains' or 'equals' parameter", environment.getFieldDefinition().getName()));
 
             if (arguments.containsKey(CONTAINS)) {
-                String argument = Text.escapeIllegalXpathSearchChars((String) arguments.get(CONTAINS));
+                String argument = JCRContentUtils.sqlEncode((String) arguments.get(CONTAINS));
```
eallais: Small behaviour change worth being deliberate about: `contains: null` used to blow up, and now it silently queries for the four-character string `null`.

- `containsKey` is true for an explicitly-null GraphQL argument, so the branch is entered with a null value. The old `Text.escapeIllegalXpathSearchChars(null)` threw an NPE; `sqlEncode` is null-safe and returns null, which `String.format` renders as `null` inside the quotes — giving `n.[prop]='null'`.
- Neither behaviour is good, and the new one is arguably worse for being quiet: an NPE surfaces, a query that matches nodes whose property literally reads "null" does not.
- Same shape on both branches here and on the two in `WeakreferenceFinderDataFetcher`.

Cheapest fix is to treat an explicitly-null argument like a missing one and let the existing "must have either 'contains' or 'equals'" error carry it. Happy to leave it if you'd rather keep the diff to the encoder swap — just don't want it to land unnoticed.

**Source:** `jahia-private core/src/main/java/org/jahia/services/content/JCRContentUtils.java:1787` — `sqlEncode` returns its argument untouched when it is null.

graphql-extension-example/src/main/resources/META-INF/graphql-extension.sdl
```diff
@@ -24,6 +24,8 @@ extend type Query {
     newsByChecked: [NewsSDL]
     """This is myNewsByDate"""
     myNewsByDate: [NewsSDL]
+    """This is myNewsByTitle"""
+    myNewsByTitle: [NewsSDL]
```
eallais, line 28: I put the String finder in `graphql-extension-example` because it already declares `myNewsByDate` next door and is already deployed in the integration environment — but I want to check that is the right home.

- This module doubles as reference material, and its SDL is the thing people copy when they add a finder. A field that exists only to give a test something to query is a slightly odd thing to leave in an example.
- `tests/jahia-module/sdl-test-module` is the alternative and is unambiguously a test artifact — but `jnt:offer` there has no Date property, so covering the date path would mean touching its `.cnd` too.

Fine either way as far as I'm concerned; flagging it so the choice is made rather than inherited from whichever module was easier to reach.

tests/cypress/fixtures/groovy/prepareSDLFinderLiteralTest.groovy
eallais: Fixed in `ee81b257`: the fixture now sets `date` (not i18n) from the plain system session, then reopens `doExecuteWithSystemSessionAsUser(root, "default", Locale.ENGLISH, …)` for `jcr:title` and `desc` — the same split `prepareValidityTest.groovy` uses.

**Not verified yet, so leaving this thread open.** The spec still has not run: `graphql-extension-example` declares `Jahia-Depends: default,news,graphql-dxm-provider`, my lab has no `news` module and I could not fetch `news:3.4.0` from here, so the finder is absent from the schema and there is nothing to query. The repo's own integration job is red for an unrelated environment reason (same failure signature as #632, `0 passing` across specs nobody touched). Claiming this fixed on a compile would repeat exactly what the other thread flags.

_(the branch was rebuilt into a single commit, so the sha above replaces the one this reply first cited.)_

tests/cypress/e2e/sdl/sdlFinderLiterals.cy.ts
eallais: Fixed in `ee81b257`: each negative case now queries the plain value first and requires the marker back before firing the quote-carrying one — `equals: "sdlLit plain marker"` must return the node, `contains: "plain marker"` must include it, and the date case pins the in-range set before asserting the 1999 node stays out. A fixture that produces nothing now fails the case meant to catch it.

Also extracted `queryTitles`/`queryDates` so the anchor and the quoted value read side by side.

**Not verified yet, so leaving this thread open** — same reason as the fixture thread: no `news` module on the lab, and the integration job is red for an environment reason unrelated to this PR.

_(the branch was rebuilt into a single commit, so the sha above replaces the one this reply first cited.)_

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/sdl/fetchers/StringFinderDataFetcher.java
eallais: Fixed in `ee81b257`: both fetchers read `contains`/`equals` before branching, so an explicitly-null argument no longer reaches `String.format`. `contains: null` with no `equals` now hits the existing "must have either 'contains' or 'equals'" error instead of querying for the literal string `null`; `contains: null` alongside a real `equals` falls through to the equals branch, which is what the caller meant.

Same change on both branches of `WeakreferenceFinderDataFetcher`. Compiles green; this one is plain Java with no runtime dependency on the test harness, so resolving.

_(the branch was rebuilt into a single commit, so the sha above replaces the one this reply first cited.)_

graphql-extension-example/src/main/resources/META-INF/graphql-extension.sdl
eallais: Keeping it in `graphql-extension-example`, resolving as no change.

`myNewsByDate` is already declared here and the module is already deployed in the integration environment, so the String finder sits next to its sibling and costs nothing extra to reach. Moving it to `sdl-test-module` would mean adding a Date property to `jnt:offer`'s `.cnd` purely to cover the date path, and rewriting fixture and spec around it — more moving parts in the test surface, for a field that reads naturally beside the one already there.

Worth revisiting if the example module ever stops being deployed for the integration run.

tests/cypress/fixtures/groovy/prepareSDLFinderLiteralTest.groovy
eallais: Fixed in the branch and now proven: the fixture writes `jcr:title` from a locale-bearing session, and the spec reads it back with the field's own `language` argument (`1d34ab61`) — an i18n field needs both, the finder's language only sets the query session's locale. Integration tests pass, all five cases green.

tests/cypress/e2e/sdl/sdlFinderLiterals.cy.ts
eallais: Fixed in the branch and now proven: every negative case is anchored — it first asserts the plain value does select the marker, so an empty result can no longer come from an absent fixture. Verified end to end against a running instance before pushing (5/5), then green in the integration run (`1d34ab61`).

baptistegrimaud (approved):

PR #645, opened 2026-07-28
fix: bound mutation batch size with a dedicated configurable limit

AKarmanov:
## Summary

Three mutation fields took their cardinality from the caller with no upper bound:

- **`mutateNodesByQuery`** — its `limit` argument is optional, so with none supplied every matching node was
  materialized as a `GqlJcrNodeMutation`.
- **`mutateNodes(pathsOrIds:)`** and **`addNodesBatch(nodes:)`** — iterate whatever collection they are handed.

The pre-existing `graphql.fields.node.limit` did not cover them: it is applied by `PaginationHelper`, so it reaches
relay `@GraphQLConnection` fields only, while these build a plain `Collection`. `nodesByQuery` was therefore bounded and
its mutation counterpart was not, even though the shipped configuration described result-set size as bounded by that
property.

## Change

A dedicated property, **`graphql.mutation.batch.limit`** (default `5000`, `0` disables), read only from the default
configuration file like the other limits so a non-default configuration cannot loosen it. The rules live in
`GraphQLLimits`, fed by `DXGraphQLConfig`.

**Why a separate property rather than reusing `graphql.fields.node.limit`:** that one caps how many nodes a read field
collects and serializes; a mutation batch accumulates pending changes in one JCR session written in a single commit, so
an operator should be able to tighten one without shrinking the other. It is also distinct from the query-cost guards,
which measure the shape of a document rather than how many items its arguments carry. Keeping it separate leaves
`PaginationHelper` — whose accessors are exported — untouched.

**Where the bound is applied follows from what is knowable before execution:**

- An **enumerated** batch states its size in its arguments, so `QueryCostCalculator` measures it as a third metric of
  the single document walk it already performs, and `QueryCostInstrumentation` refuses the request before any field is
  fetched, permission-checked or serialized. The total is summed across the whole document and nested input objects are
  followed, so it reflects what the request asks for however it is expressed. Refusing rather than truncating is
  deliberate: the caller named those nodes, so operating on a prefix of the list would quietly not do what was asked.
- **`mutateNodesByQuery`** cannot be measured that way, since how many nodes a JCR-SQL2 statement matches is only known
  once it runs. It is bounded when the query executes and **truncates** instead, drawing from what the request has left
  rather than from the whole allowance, and logs when it reaches the bound.

Applied to **mutations only**: a query handed a list of paths to read is bounded per connection at execution time, and a
write-sized budget would reject reads that are cheap today.

### Behaviour change

A request operating on **more than 5000 nodes at once** is now refused (enumerated) or truncated (query-driven) and must
be split — with `limit`/`offset` for `mutateNodesByQuery`, or by raising the property. The default matches the
pre-existing node limit, so ordinary traffic is unchanged and introducing the dial is not itself a change of behaviour.
It is a starting point rather than a considered figure: a write costs more per node than a read, so it is expected to be
lowered once there is data on real batch sizes.

Note that existing installs keep the bound via the code default — `fileinstall` seeds a `-default.cfg` on first install
only, so the shipped value does not reach them, but `GraphQLLimits`' own default is the same 5000. Only *tuning*
requires editing the configuration.

## Verification

- **Unit** (`QueryCostCalculatorTest`, no instance needed): the items an enumerated argument carries, the same items via
  a variable, items nested in a recursive input type, scalars and argument-less fields contributing nothing, and
  aliased selections summing rather than resetting.
- **Integration** (`GraphQLNodeMutationsTest`): the query-driven bound and its interaction with an explicit `limit`,
  both enumerated forms refused, aliases sharing one allowance in both forms, and `0` disabling the bound.
- **Cypress** (`mutationBatchLimit.cy.ts`): the same bound over HTTP with the list supplied through a variable,
  including a batch nested in a recursive input type, plus a within-bound request still succeeding.

All three levels were run against a local instance before pushing, and a read selecting 5001 paths was confirmed
unaffected.


.chachalog/J8wXp.md
```diff
@@ -0,0 +1,9 @@
...
+
+**Are you affected?** Only if a single request operates on more than 5000 nodes — the default matches the existing node limit, so ordinary traffic is unchanged. If it does, either split the work into smaller requests (using `limit`/`offset` for `mutateNodesByQuery`), or raise `graphql.mutation.batch.limit` in `org.jahia.modules.graphql.provider-default.cfg`.
```
gflores-jahia: Reads too verbose for a changelog imo. First sentence I think captures the change already. **Are you affected** is nice to have that we can also keep. Otherwise, the rest sounds implem details that needs to go either in the PR or issue desc if it isn't already.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/instrumentation/QueryCostCalculator.java
```diff
@@ -125,14 +146,54 @@ private static final class Measurement {
...
+         * the outermost list. Nesting has to be followed because an input type can hold a list of itself - {@code
+         * InputJCRNode.children} is a list of {@code InputJCRNode} - so one outer item can describe an arbitrarily large
+         * tree of nodes to create, and counting only the outer list would score that as a single item however much work
+         * it asks for.
+         */
+        private static int countItems(Object value) {
+            if (value instanceof Collection) {
+                int items = ((Collection<?>) value).size();
+                for (Object element : (Collection<?>) value) {
+                    items += countItems(element);
```
gflores-jahia: flagged possible stack overflow with this recursion. Suggested fix:

> Rewrite countItems the same way depthOf was written: an explicit Deque-based worklist instead of recursion, and — since there's no reason to keep counting once the answer is already "reject" — bail out early once the running total exceeds maxBatchSize (which requires either passing the limit down into the counting pass, or checking it incrementally in enumeratedItems/measure rather than only once at the end in QueryCostInstrumentation). That removes both the stack-depth risk and the wasted full-structure walk on oversized payloads, using the same style the author already applied to depthOf right next to it.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/instrumentation/QueryCostInstrumentation.java
```diff
@@ -45,22 +50,28 @@ public class QueryCostInstrumentation extends SimplePerformantInstrumentation {
...
     @Override
     public InstrumentationContext<ExecutionResult> beginExecuteOperation(InstrumentationExecuteOperationParameters parameters, InstrumentationState state) {
+        ExecutionContext executionContext = parameters.getExecutionContext();
         QueryCostCalculator.QueryCost cost =
-                QueryCostCalculator.calculate(QueryCostCalculator.newTraverser(parameters.getExecutionContext()));
+                QueryCostCalculator.calculate(QueryCostCalculator.newTraverser(executionContext));
         if (logger.isDebugEnabled()) {
-            logger.debug("Query cost: complexity {}, depth {}", cost.getComplexity(), cost.getDepth());
+            logger.debug("Query cost: complexity {}, depth {}, batch size {}",
+                    cost.getComplexity(), cost.getDepth(), cost.getBatchSize());
```
gflores-jahia, line 76: optimization: batch size only checked when isMutation(executionContext) is true but is calculated everytime. maybe move isMutation check prior to the calculation of batch size.

gflores-jahia (commented):
Note: I added AI findings verbatim for full context, but I added code-specific comments for targeted fixes (at least for the ones I think that matter) but maybe AI can create fixes for all issues.

----

## Findings

### 1. (High) The pre-execution guard sums *every* list argument in the mutation document, not just node-batch arguments — real false-positive rejection risk

`QueryCostCalculator.enumeratedItems`/`countItems` walks every argument of every field in a
mutation and recurses into nested input objects, with no filtering by field/argument identity.
But plenty of mutation fields take lists that have nothing to do with "how many nodes this
request touches":

- `GqlJcrPropertyMutation.setValues(values: List<String>)` (line 111) — setting one multivalued
  property to N values now counts N toward the *node*-batch ceiling.
- `GqlJcrNodeMutation.reorderChildren(names: List<String>)` (line 498) — reordering all children
  of a large listing.
- `grantRoles`/`revokeRoles(roleNames: List<String>)` (541/552), `addMixins`/`removeMixins(mixins)`
  (304/324).
- `setPropertiesBatch(properties: Collection<GqlJcrPropertyInput>)` (291) double-counts: the outer
  list *and* each entry's nested `values` list.

A legitimate mutation that sets a large multivalued property (tags, category lists, translated-
language lists) or reorders a big folder can now hit
`AbortExecutionException("maximum mutation batch size exceeded")` with zero relation to "nodes
operated on." The chachalog's "are you affected?" note ("only if a single request operates on
more than 5000 nodes") is therefore not accurate — it should read "more than 5000 *list-argument
items* anywhere in the mutation," which is a materially different and harder-to-predict claim.

### 2. (Medium) `countItems` recurses with no depth bound, unlike the two existing metrics

`maxQueryDepth` bounds *selection* nesting, not *input-value* nesting. A narrow-but-deep
`InputJCRNode.children` chain (or any recursive input) drives unbounded recursion in
`countItems` before any permission check runs, before the size cutoff even matters (it doesn't
short-circuit once the running total exceeds the limit) — a StackOverflowError vector under
whatever depth an upstream JSON parser tolerates. Cheap fix: an explicit depth cap, or convert to
iterative (Deque-based, matching the existing traversal style) traversal with an early-exit once
the count already exceeds the limit.

### 3. (Low/Medium, easy fix) The new per-field work runs for every request, including plain queries, then is discarded

`QueryCostInstrumentation` only *applies* the batch check when `isMutation(executionContext)`,
but `QueryCostCalculator.calculate()` — including the new `enumeratedItems` recursion — runs
unconditionally whenever any of the three guards is enabled (true for shipped defaults).
Read-heavy traffic (jContent listings, `children(names: [...])`, `byPath(paths: [...])`) now pays
for counting list-argument cardinality it will never use. Fix: thread the operation type into
`Measurement` and skip the batch-size accumulation branch entirely for query operations.

### 4. (Medium) Silent truncation on writes is a worse failure mode than on reads

The fiche's own example — `content-releases`' `RemoveRelease` calling `mutateNodesByQuery` to
strip references — is exactly the kind of caller that could exceed 5000 matches. Previously it
always completed; now it silently truncates with only a server-side WARN log, no client-visible
signal. A caller doing "strip all references" and getting a truncated result set with no
indication in the GraphQL response can leave data half-migrated with no error raised. Consider
surfacing truncation to the caller (a boolean/`hasMore` field on the mutation result, or a
GraphQL `extensions` entry) rather than only a server log.

### 5. (Minor) The WARN log fires even when there was no actual truncation

It fires whenever `result.size() == effectiveLimit`, even when the query happened to match
exactly that many nodes (no truncation actually occurred) — "additional matching nodes, if any,
were not returned" is then misleading. Not a security/perf issue, just log noise.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/instrumentation/QueryCostCalculator.java
```diff
@@ -143,57 +158,82 @@ int getBatchSize() {
...
-                items += countItems(argument);
+            for (Map.Entry<String, Object> argument : env.getArguments().entrySet()) {
+                GraphQLArgument definition = field.getArgument(argument.getKey());
+                if (definition == null) {
+                    continue;
+                }
+                String type = GraphQLTypeUtil.unwrapAll(definition.getType()).getName();
+                if (NODE_INPUT_TYPES.contains(type)) {
+                    items += countNodes(argument.getValue());
+                } else if (NODE_PATHS_ARGUMENT.equals(argument.getKey()) && argument.getValue() instanceof Collection) {
```
gflores-jahia: thanks for the added checks but looks we still have a gap when the type check fails and we fall back on checking NODE_PATHS_ARGUMENT. Not all that uses pathsOrIds belong to node mutation e.g. `ZipFileMutation.addToZip(pathsOrIds: [String]!)` uses same argument name and counts as part of limit but does not modify these nodes.

Possible check to add would be to require both the field name and argument to avoid false positives, in this case I think we're only matching mutateNodes, maybe something like:

```
private static final String NODE_PATHS_FIELD = "mutateNodes";
[...]
if (NODE_INPUT_TYPES.contains(type)) {
                    items += countNodes(argument.getValue());
                } else if (NODE_PATHS_FIELD.equals(field.getName()) && NODE_PATHS_ARGUMENT.equals(argument.getKey())
                        && argument.getValue() instanceof Collection) {
                    items += ((Collection<?>) argument.getValue()).size();
                }
```

gflores-jahia (commented):
found another possible improvement to the checks

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/instrumentation/QueryCostCalculator.java
AKarmanov: Agreed on `addToZip` — it takes the same argument name to write a single archive, so its cardinality is one however many paths it is handed and it has no batch to size. Fixed in b158281b.

I went with a slightly different discriminator than the suggestion, because three fields declare `pathsOrIds`, not one:

| field | returns | batches over the list |
|---|---|---|
| `GqlJcrMutation.mutateNodes` | `Collection<GqlJcrNodeMutation>` | yes |
| `VanityUrlJCRMutationExtensions.mutateVanityUrls` | `Collection<GqlVanityUrlMappingMutation>` | yes |
| `GqlZipMutation.addToZip` | `boolean` | no |

Pinning the field name to `mutateNodes` would drop `mutateVanityUrls`, and that one has nothing else holding it: `mutateNodes` calls `GraphQLLimits.checkMutationBatchSize()` itself in the resolver, `mutateVanityUrls` is sized only here. A name list is also closed to the batch mutations extension providers add to this schema, which are named in no list this class could hold.

So the match is on the **pair** — a `pathsOrIds` list *and* a list return type — which is the property the two real batches share and the archive field does not:

```java
private boolean isNodeBatch(GraphQLFieldDefinition field, String argumentName) {
    return NODE_PATHS_ARGUMENT.equals(argumentName)
            && GraphQLTypeUtil.isList(GraphQLTypeUtil.unwrapNonNull(field.getType()));
}
```

Three tests in `QueryCostCalculatorTest`: 4000 paths through `addToZip` scoring 0 (fails without the change, `expected:<0> but was:<4000>`), `mutateVanityUrls` counted at its own size, and the two batch fields sharing one allowance. The last two are there as guards against the name-based variant — they pass today either way, and fail if the match is ever narrowed to a field name.

The `pathsOrIds` naming convention is still what carries this, so a batch mutation that names its argument something else escapes the pre-execution measure. That was already the case, and the resolver-side check in `mutateNodes` is the backstop; worth a follow-up to give the other batch fields the same, if you think it is warranted.

gflores-jahia (approved):

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/instrumentation/QueryCostCalculator.java
gflores-jahia: this is a better solution, thanks

PR #650, opened 2026-08-04
fix: scope Tag Manager mutations to the nodes the caller may write

eallais:
## Summary

The Tag Manager mutations (`renameTag`, `deleteTag`, `renameTagOnNode`, `deleteTagOnNode`) perform
their writes on a **system session**, which is needed because each operation updates the `default`
and `live` copies in one pass and the caller is not expected to hold live write rights. The
consequence is that the only right consulted is the site-level `tagManager` permission that opens
the screen: once past it, the tag list of **any** node under the site is rewritten, including nodes
the caller has no rights on. That contradicts the JCR permission model — a site-level permission is
not meant to widen what a caller may write inside the site.

## Change

`TagManagerMutationService` now resolves each candidate node in the **caller's own** session before
touching it, and only then performs the write on the system session:

- New `isModifiableByCaller(...)`: resolves the node by identifier in the caller's `default`-workspace
  session and checks `jcr:modifyProperties`. The edit workspace is the authority for both workspaces,
  since the live copy is updated by the same operation. A node the caller cannot resolve there at all
  (not readable, or no longer present in the edit workspace) is reported as not modifiable — the
  fail-closed answer. The verdict depends only on the identifier and the caller's session, so it is
  computed once per node and reused by both workspace passes. A node the caller's session cannot
  resolve is logged at debug level: resolving an identifier walks every store provider and reports
  an access decision and a store failure alike, so the reason has to be recoverable from the logs
  rather than left as an unexplained entry in the failure count.
- The two **site-wide** operations share one body (`applyUnderSite`), which iterates the tagged nodes
  per workspace and applies the check per node. A node the caller may not write is left untouched and
  counted as a failure, which the result type already models (`failedCount` is documented as possibly
  exceeding `failedPaths`); its path is deliberately not reported, since the caller may not be able to
  read it either. Per-node delegation is to `TaggingService.renameTag` / `untag` — the same primitives
  the single-node mutations already use.
- The candidate query is the **parameter-bound** form already used on the read side, rather than the
  concatenated one. For a tag carrying a quote the two forms do not select the same nodes, so
  `taggedContent` and the mutations now agree on the candidate set — pinned by a spec case.
- The two **single-node** operations gain `ensureCallerCanModify(...)` right after the existing
  site-membership validation, so an out-of-reach node is refused instead of written.
- `ensureMutationTagName` additionally refuses a replacement name that the configured `TagHandler`
  reduces to nothing. Driving the bulk rename node by node drops the guard
  `TaggingService.renameTagUnderPath` applied, and a sanitizing handler could otherwise store an
  empty tag value; placing the check on the argument covers the single-node mutations too, which
  never had it.

Behavior on the normal path is unchanged: a caller with `tagManager` and write rights over the site
processes exactly the same nodes as before, in both workspaces, with the same batching, observation
suppression and partial-failure semantics.

**One behavior change worth naming.** Because the verdict always resolves the node in the `default`
workspace, a node that exists only in `live` — published, then removed from the edit workspace — is
reported as not modifiable and keeps its tag, where the previous site-wide path cleaned those stale
live tags on the way through. That is deliberate: the check is fail-closed, and no caller is expected
to hold live write rights. Such a node is counted in `failedCount` without its path.

## Verification

Extended `tests/cypress/e2e/api/tagManager.cy.ts` with a `per-node write rights` block: a user
holding `server-administrator` + `site-administrator` (so `tagManager` on the site is granted) and a
node under that site where both roles are denied, published beforehand so the tag exists in both
workspaces. A separate case covers the quoted tag on the site-wide rename.

Run against a Jahia 8.2.4.0-SNAPSHOT lab with the module deployed:

| build | result |
|---|---|
| before this change | 21 passing, **4 failing** — the site-wide rename reports `processedCount` 2 (both workspaces written on the out-of-reach node) and neither single-node mutation raises an error |
| with this change | **26 passing** — the out-of-reach node keeps its tag in both workspaces, the two single-node mutations are refused, the quoted tag is selected and renamed, and the positive control (renaming a tag the caller may write) still processes its node |

The 20 pre-existing tests in the spec pass unchanged on both builds. The new rights block includes
that positive control precisely so a check that merely stopped writing everything could not pass it.
The quoted-tag case pins the intended selection on this build; it was not run against the previous
concatenated form.

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
@@ -21,6 +21,7 @@ import org.jahia.modules.graphql.provider.dxm.DataFetchingException;
 import org.jahia.modules.graphql.provider.dxm.node.GqlJcrWrongInputException;
 import org.jahia.modules.graphql.provider.dxm.service.tags.graphql.GqlTagMutationResult;
 import org.jahia.modules.graphql.provider.dxm.service.tags.graphql.GqlTagWorkspaceMutationResult;
+import org.jahia.services.content.JCRContentUtils;
 import org.jahia.services.content.JCRNodeWrapper;
 import org.jahia.services.content.JCRObservationManager;
 import org.jahia.services.content.JCRSessionFactory;
@@ -30,7 +31,9 @@ import org.jahia.services.tags.TaggingService;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 
+import javax.jcr.NodeIterator;
 import javax.jcr.RepositoryException;
+import javax.jcr.query.Query;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -55,7 +58,9 @@ import java.util.List;
  *       per-node errors and continue processing; callers receive a structured result describing
  *       which nodes succeeded and which failed, without a full transaction rollback.</li>
  *   <li><strong>Authorization</strong> – every operation checks that the current user holds the
- *       {@code tagManager} permission on the target site node before performing any write.</li>
+ *       {@code tagManager} permission on the target site node before performing any write, and each
+ *       individual node is additionally checked against the caller's own rights on it — the
+ *       site-level permission opens the screen, it does not widen what the caller may write.</li>
  * </ul>
  *
  * <p><strong>Threading model:</strong> this is a singleton OSGi component. Each public method
@@ -70,6 +75,9 @@ import java.util.List;
 public class TagManagerMutationService {
     private static final List<String> WORKSPACES = Arrays.asList(Constants.EDIT_WORKSPACE, Constants.LIVE_WORKSPACE);
 
+    /** Right the caller must hold on a node for its tag list to be rewritten. */
+    private static final String MODIFY_PROPERTIES_PERMISSION = "jcr:modifyProperties";
+
     /**
      * Maximum number of failure paths included in a single response payload.
      * Additional failures are counted but not listed.
@@ -83,11 +91,12 @@ public class TagManagerMutationService {
      * Renames a tag across <em>all</em> nodes in the site that carry it, in both the
      * {@code default} and {@code live} workspaces.
      *
-     * <p>Delegates to {@link TaggingService#renameTagUnderPath} for each workspace, which handles
-     * node iteration and per-node post-processing via {@link TagManagerActionCallback}. JCR
-     * observation listeners are disabled for the entire batch and restored in a {@code finally}
-     * block. Individual node failures do not abort the batch — they are captured in the
-     * returned result with partial-failure semantics.
+     * <p>Iterates the tagged nodes of each workspace and delegates the per-node rewrite to
+     * {@link TaggingService#renameTag(JCRNodeWrapper, String, String)}, with post-processing via
+     * {@link TagManagerActionCallback}. A node the caller may not write is left untouched and
+     * counted as a failure. JCR observation listeners are disabled for the entire batch and
+     * restored in a {@code finally} block. Individual node failures do not abort the batch — they
+     * are captured in the returned result with partial-failure semantics.
      *
      * @param siteKey the Jahia site identifier (e.g. {@code "digitall"}); must not be
      *                {@code null}; authorization is pre-validated by the GraphQL resolver
@@ -103,15 +112,51 @@ public class TagManagerMutationService {
      */
     public GqlTagMutationResult renameTag(String siteKey, String tag, String newName) {
         ensureMutationTagName(newName);
+        return applyUnderSite(siteKey, tag, newName);
+    }
+
+    /**
+     * Removes a tag from <em>all</em> nodes in the site that carry it, in both the
+     * {@code default} and {@code live} workspaces.
+     *
+     * <p>Iterates the tagged nodes of each workspace and delegates the per-node removal to
+     * {@link TaggingService#untag(JCRNodeWrapper, String)}. A node the caller may not write is left
+     * untouched and counted as a failure. Observation listeners are suppressed across both
+     * workspace operations with the same guarantees as {@link #renameTag}. Individual node failures
+     * are captured with partial-failure semantics.
+     *
+     * @param siteKey the Jahia site identifier; must not be {@code null}; authorization is
+     *                pre-validated by the GraphQL resolver
+     * @param tag     the tag value to remove; must not be {@code null} or empty
+     * @return a {@link GqlTagMutationResult} carrying the tag name, a {@code null} node
+     *         identifier (bulk mode), and one {@link GqlTagWorkspaceMutationResult} per workspace
+     * @throws DataFetchingException wrapping a {@link RepositoryException} if session
+     *                               acquisition or query execution fails
+     */
+    public GqlTagMutationResult deleteTag(String siteKey, String tag) {
+        return applyUnderSite(siteKey, tag, null);
+    }
+
+    /**
+     * Shared body of the two site-wide operations: rename when {@code newName} is non-{@code null},
+     * removal otherwise.
+     *
+     * <p>The writes stay on a system session — the {@code live} copy is updated in the same pass, and
+     * the caller is not expected to hold live write rights — so every candidate node is checked
+     * against the caller's own view of it before it is touched (see
+     * {@link #isModifiableByCaller(JCRSessionWrapper, String)}).
+     */
+    private GqlTagMutationResult applyUnderSite(String siteKey, String tag, String newName) {
         String sitePath = "/sites/" + siteKey;
         try {
+            JCRSessionWrapper callerSession = JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE);
             List<GqlTagWorkspaceMutationResult> results = new ArrayList<>();
 
             JCRObservationManager.setAllEventListenersDisabled(Boolean.TRUE);
             try {
                 for (String workspace : WORKSPACES) {
                     JCRSessionWrapper systemSession = JCRSessionFactory.getInstance().getCurrentSystemSession(workspace, null, null);
-                    results.add(taggingService.renameTagUnderPath(sitePath, systemSession, tag, newName, new TagManagerActionCallback(systemSession, workspace)));
+                    results.add(applyInWorkspace(systemSession, callerSession, workspace, sitePath, tag, newName));
                 }
             } finally {
                 JCRObservationManager.setAllEventListenersDisabled(Boolean.FALSE);
@@ -123,41 +168,36 @@ public class TagManagerMutationService {
         }
     }
 
-    /**
-     * Removes a tag from <em>all</em> nodes in the site that carry it, in both the
-     * {@code default} and {@code live} workspaces.
-     *
-     * <p>Delegates to {@link TaggingService#deleteTagUnderPath} for each workspace. Observation
-     * listeners are suppressed across both workspace operations with the same guarantees as
-     * {@link #renameTag}. Individual node failures are captured with partial-failure semantics.
-     *
-     * @param siteKey the Jahia site identifier; must not be {@code null}; authorization is
-     *                pre-validated by the GraphQL resolver
-     * @param tag     the tag value to remove; must not be {@code null} or empty
-     * @return a {@link GqlTagMutationResult} carrying the tag name, a {@code null} node
-     *         identifier (bulk mode), and one {@link GqlTagWorkspaceMutationResult} per workspace
-     * @throws DataFetchingException wrapping a {@link RepositoryException} if session
-     *                               acquisition or query execution fails
-     */
-    public GqlTagMutationResult deleteTag(String siteKey, String tag) {
-        String sitePath = "/sites/" + siteKey;
-        try {
-            List<GqlTagWorkspaceMutationResult> results = new ArrayList<>();
-
-            JCRObservationManager.setAllEventListenersDisabled(Boolean.TRUE);
+    private GqlTagWorkspaceMutationResult applyInWorkspace(JCRSessionWrapper systemSession, JCRSessionWrapper callerSession,
+                                                           String workspace, String sitePath, String tag, String newName) throws RepositoryException {
+        TagManagerActionCallback callback = new TagManagerActionCallback(systemSession, workspace);
+        NodeIterator taggedNodes = queryTaggedNodes(systemSession, sitePath, tag);
+        while (taggedNodes.hasNext()) {
+            JCRNodeWrapper node = (JCRNodeWrapper) taggedNodes.nextNode();
             try {
-                for (String workspace : WORKSPACES) {
-                    JCRSessionWrapper systemSession = JCRSessionFactory.getInstance().getCurrentSystemSession(workspace, null, null);
-                    results.add(taggingService.deleteTagUnderPath(sitePath, systemSession, tag, new TagManagerActionCallback(systemSession, workspace)));
+                if (!isModifiableByCaller(callerSession, node.getIdentifier())) {
+                    callback.onSkipped();
+                    continue;
                 }
-            } finally {
-                JCRObservationManager.setAllEventListenersDisabled(Boolean.FALSE);
+                if (newName != null) {
+                    taggingService.renameTag(node, tag, newName);
+                } else {
+                    taggingService.untag(node, tag);
+                }
+                callback.afterTagAction(node);
+            } catch (RepositoryException e) {
+                callback.onError(node, e);
             }
-
-            return new GqlTagMutationResult(tag, null, results);
-        } catch (RepositoryException e) {
-            throw new DataFetchingException(e);
         }
+        return callback.end();
+    }
+
+    private NodeIterator queryTaggedNodes(JCRSessionWrapper session, String sitePath, String tag) throws RepositoryException {
+        String statement = "SELECT * FROM [jmix:tagged] AS result WHERE ISDESCENDANTNODE(result, '" +
+                JCRContentUtils.sqlEncode(sitePath) + "') AND (result.[j:tagList] = $tag)";
+        Query query = session.getWorkspace().getQueryManager().createQuery(statement, Query.JCR_SQL2);
+        query.bindValue("tag", session.getValueFactory().createValue(tag));
+        return query.execute().getNodes();
     }
 
     /**
@@ -187,6 +227,7 @@ public class TagManagerMutationService {
         try {
             JCRSessionWrapper editSession = JCRSessionFactory.getInstance().getCurrentSystemSession(Constants.EDIT_WORKSPACE, null, null);
             validateNodeBelongsToSite(editSession, nodeId, sitePath);
+            ensureCallerCanModify(nodeId);
             List<GqlTagWorkspaceMutationResult> workspaceResults = new ArrayList<>();
 
             JCRObservationManager.setAllEventListenersDisabled(Boolean.TRUE);
@@ -251,6 +292,7 @@ public class TagManagerMutationService {
         try {
             JCRSessionWrapper editSession = JCRSessionFactory.getInstance().getCurrentSystemSession(Constants.EDIT_WORKSPACE, null, null);
             validateNodeBelongsToSite(editSession, nodeId, sitePath);
+            ensureCallerCanModify(nodeId);
             List<GqlTagWorkspaceMutationResult> workspaceResults = new ArrayList<>();
 
             JCRObservationManager.setAllEventListenersDisabled(Boolean.TRUE);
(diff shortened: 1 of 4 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -292,6 +334,37 @@ static void flushNodeCaches(String path) {
...
+     * the same operation to keep the two in sync, and a caller who may edit content is not expected to
+     * hold live write rights. A node the caller cannot resolve there at all — unreadable, or no longer
+     * present in the edit workspace — is reported as not modifiable.
+     */
+    private boolean isModifiableByCaller(JCRSessionWrapper callerSession, String nodeIdentifier) {
+        try {
+            return callerSession.getNodeByIdentifier(nodeIdentifier).hasPermission(MODIFY_PROPERTIES_PERMISSION);
+        } catch (RepositoryException e) {
+            return false;
+        }
```
eallais, line 400: This catch turns *any* repository failure into "not modifiable", and nothing anywhere records that it happened.

- A transient `RepositoryException` (a broken session, a store hiccup) is indistinguishable from a genuine access decision — the node is skipped, `failedCount` goes up, and the path is deliberately not reported.
- The class has no logger, so on the bulk path the result a support engineer gets is "42 processed, 3 failed" with no way to learn which nodes or why. `onSkipped()` has the same property by design, which is fine for a real access decision but not for an error.
- Worth narrowing the catch to the access-denied cases and letting the rest surface, or at minimum adding a `logger.debug` on the deny branch with the identifier and the exception.

**The gate is now the thing that decides whether a write happens — it deserves to be diagnosable.**

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -160,6 +168,38 @@ public GqlTagMutationResult deleteTag(String siteKey, String tag) {
...
 
+    private GqlTagWorkspaceMutationResult applyInWorkspace(JCRSessionWrapper systemSession, JCRSessionWrapper callerSession,
+                                                           String workspace, String sitePath, String tag, String newName) throws RepositoryException {
+        TagManagerActionCallback callback = new TagManagerActionCallback(systemSession, workspace);
+        NodeIterator taggedNodes = queryTaggedNodes(systemSession, sitePath, tag);
+        while (taggedNodes.hasNext()) {
+            JCRNodeWrapper node = (JCRNodeWrapper) taggedNodes.nextNode();
+            try {
+                if (!isModifiableByCaller(callerSession, node.getIdentifier())) {
+                    callback.onSkipped();
```
eallais, line 203: Because the verdict always resolves the node in the `default` workspace, a live node that no longer exists there can never be touched again.

- The live pass iterates the live copies, but the check resolves each identifier in `default`; for a node that was published and then removed from the edit workspace, `getNodeByIdentifier` misses and the answer is a permanent "not modifiable".
- Before this change the site-wide operation cleaned those stale live tags. Now they stay, counted as an unexplained failure with no path — so the operator can neither fix them through this API nor find out which they are.
- Fail-closed is the right default here, and no caller holds live write rights anyway, so I don't think the check should change. But it is a behavior change worth naming — either in the PR body or, better, by reporting those particular skips with their path, since the caller *can* read them in live.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -160,6 +168,38 @@ public GqlTagMutationResult deleteTag(String siteKey, String tag) {
...
     }
 
+    private GqlTagWorkspaceMutationResult applyInWorkspace(JCRSessionWrapper systemSession, JCRSessionWrapper callerSession,
+                                                           String workspace, String sitePath, String tag, String newName) throws RepositoryException {
+        TagManagerActionCallback callback = new TagManagerActionCallback(systemSession, workspace);
+        NodeIterator taggedNodes = queryTaggedNodes(systemSession, sitePath, tag);
+        while (taggedNodes.hasNext()) {
+            JCRNodeWrapper node = (JCRNodeWrapper) taggedNodes.nextNode();
+            try {
+                if (!isModifiableByCaller(callerSession, node.getIdentifier())) {
```
eallais: `applyInWorkspace` runs once per workspace, so each node's verdict is computed twice — from the same session, on the same identifier, with the same answer by construction.

- That is `2N` extra `getNodeByIdentifier` + `hasPermission` round-trips on the caller session for a site with `N` tagged nodes, which is exactly the cost this API pays on the tag a site actually uses everywhere.
- Since the `default` workspace is the stated authority for both passes, the set of writable identifiers could be resolved once in `applyUnderSite` and handed to both workspace passes — or simply memoized in a `Set<String>` / `Map<String, Boolean>` across the loop.

**Not a correctness issue, just work done twice on the path most likely to be large.**

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -103,33 +112,18 @@ public class TagManagerMutationService {
      */
     public GqlTagMutationResult renameTag(String siteKey, String tag, String newName) {
         ensureMutationTagName(newName);
```
eallais, line 134: Driving the bulk rename directly drops a guard `renameTagUnderPath` used to apply: it only proceeded when `tagHandler.execute(tagNewName)` was non-empty.

- With the stock `TagHandlerImpl` (trim + lowercase) `ensureMutationTagName`'s `isBlank` check already covers it, so nothing changes on a default install.
- With a custom `TagHandler` — it is a Spring-injected strategy, so a sanitizing implementation is legitimate — a name that normalizes to empty used to make the bulk rename a no-op. Now it reaches `TaggingService.renameTag`, which adds `tagHandler.execute(splitedTag)` to `j:tagList`, i.e. an empty tag value.
- Folding it into `ensureMutationTagName` would cover all four mutations at once (the single-node paths never had the guard either) and is strictly better than the old behavior, which returned `null` into the workspace-results list.

**Was dropping it deliberate, or worth pulling into `ensureMutationTagName`?**

_Source: core `TaggingService.renameTagUnderPath` gates on `StringUtils.isNotEmpty(tagHandler.execute(tagNewName))`; `renameTag` writes `tagHandler.execute(splitedTag)` into `j:tagList`; `TagHandlerImpl.execute` is `trim().toLowerCase()`._

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -160,6 +168,38 @@ public GqlTagMutationResult deleteTag(String siteKey, String tag) {
...
+            JCRNodeWrapper node = (JCRNodeWrapper) taggedNodes.nextNode();
+            try {
+                if (!isModifiableByCaller(callerSession, node.getIdentifier())) {
+                    callback.onSkipped();
+                    continue;
+                }
+                if (newName != null) {
+                    taggingService.renameTag(node, tag, newName);
+                } else {
+                    taggingService.untag(node, tag);
```
eallais, line 209: The gate checks `jcr:modifyProperties`, but on the delete path the operation can go further than a property write.

- When the removed tag was the last one, `TaggingService.untag` removes the `j:tagList` property *and* `removeMixin(jmix:tagged)` — a node-type change, which elsewhere in JCR is `jcr:nodeTypeManagement` territory.
- The write runs on the system session, so no ACL error surfaces; the check in front of it is the only thing standing there.
- Probably acceptable — a caller who may edit the tag list arguably may drop the mixin that carries it — but if you want the check to match the operation, the delete path could additionally require `jcr:nodeTypeManagement`.

_Source: core `TaggingService.untag` calls `tagsProp.remove()` then `node.removeMixin(JMIX_TAGGED)` when the last tag goes._

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -160,6 +168,38 @@ public GqlTagMutationResult deleteTag(String siteKey, String tag) {
...
+            }
+        }
+        return callback.end();
+    }
+
+    private NodeIterator queryTaggedNodes(JCRSessionWrapper session, String sitePath, String tag) throws RepositoryException {
+        String statement = "SELECT * FROM [jmix:tagged] AS result WHERE ISDESCENDANTNODE(result, '" +
+                JCRContentUtils.sqlEncode(sitePath) + "') AND (result.[j:tagList] = $tag)";
+        Query query = session.getWorkspace().getQueryManager().createQuery(statement, Query.JCR_SQL2);
+        query.bindValue("tag", session.getValueFactory().createValue(tag));
```
eallais, line 223: Binding the tag is the right way to build this query, and it aligns the mutation with `TagManagerReadService` — but it is a semantic change worth being explicit about.

- The path this replaces built the comparison by concatenation, through `Text.escapeIllegalXpathSearchChars(selectedTag).replaceAll("'", "''")`. For a tag containing `'`, `[`, `]` or `*`, the two forms do not select the same nodes.
- Reading it as a fix rather than a regression: a tag with a quote in it is now matched by the same rule the read side already uses, so `taggedContent` and `renameTag` finally agree on the candidate set.
- Since it is unmentioned in the description and untested, one case in the spec (a tag containing a quote, renamed site-wide) would pin the intended behavior.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
eallais: Won't fix — the gate stays at `jcr:modifyProperties` on both paths.

The mixin only goes away as a side effect of removing the last tag, so requiring `jcr:nodeTypeManagement` on the delete path would refuse a legitimate editor dropping their own last tag — a stricter contract than the one rewriting `j:tagList` implies. A caller who may rewrite the tag list may also drop the mixin that carries it.

Leaving the thread open for you to close.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
eallais: Fixed in ae8bcfa: the deny branch now logs the identifier and the exception at debug level.

Narrowing the catch turned out not to help — `JCRSessionWrapper.getNodeByUUID` walks every store provider and re-wraps whatever it caught as `ItemNotFoundException`, so a genuine access decision and a store hiccup reach this line with the same type. The log is the only thing that can tell them apart; the answer stays fail-closed.

Verified with the spec run: 26/26 on a 8.2.4.0-SNAPSHOT lab with the module deployed.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
eallais: Named in the PR body, under *One behavior change worth naming*: a node that exists only in `live` is reported as not modifiable and keeps its tag, counted in `failedCount` without its path.

The check itself is unchanged, and I did not add path reporting for those particular skips: separating them from a genuine deny needs a second per-node read check, for nodes no caller can act on through this API anyway. Documented rather than reported.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
eallais: Fixed in ae8bcfa: the verdict is memoized per identifier in a `Map` created by `applyUnderSite` and shared by the two workspace passes, so each node costs one `getNodeByIdentifier` + `hasPermission` instead of two.

Verified with the spec run: 26/26 on a 8.2.4.0-SNAPSHOT lab with the module deployed.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
eallais: Not deliberate — fixed in ae8bcfa by folding it into `ensureMutationTagName`, which now also refuses a name the configured `TagHandler` reduces to nothing. Being on the argument check, it covers the two single-node mutations that never had it, and it fails the call instead of returning `null` into the workspace-results list. Changelog fragment updated to name it.

Verified with the spec run: 26/26 on a 8.2.4.0-SNAPSHOT lab with the module deployed, including the existing blank-name case.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
eallais: Pinned in 5a11ada: new case *renames a tag containing a quote on all nodes that carry it* — a node tagged `o'brien`, renamed site-wide, then found again under its new name. The PR body now states the change explicitly, in the bullet on the candidate query.

Run green with the rest of the spec: 26/26 on a 8.2.4.0-SNAPSHOT lab. It pins the intended selection on this build; I did not re-run it against the previous concatenated form.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -68,8 +77,13 @@
  */
 @Component(service = TagManagerMutationService.class, immediate = true)
 public class TagManagerMutationService {
+    private static final Logger logger = LoggerFactory.getLogger(TagManagerMutationService.class);
+
     private static final List<String> WORKSPACES = Arrays.asList(Constants.EDIT_WORKSPACE, Constants.LIVE_WORKSPACE);
 
+    /** Right the caller must hold on a node for its tag list to be rewritten. */
+    private static final String MODIFY_PROPERTIES_PERMISSION = "jcr:modifyProperties";
```
baptistegrimaud: minor: can't we use `javax.jcr.security.Privilege#JCR_MODIFY_PROPERTIES` instead?

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -292,6 +344,42 @@ static void flushNodeCaches(String path) {
...
+     *
+     * <p>The {@code default} workspace is the authority for both workspaces: the live copy is updated by
+     * the same operation to keep the two in sync, and a caller who may edit content is not expected to
+     * hold live write rights. A node the caller cannot resolve there at all — unreadable, or no longer
+     * present in the edit workspace — is reported as not modifiable.
+     */
+    private boolean isModifiableByCaller(JCRSessionWrapper callerSession, String nodeIdentifier) {
+        try {
+            return callerSession.getNodeByIdentifier(nodeIdentifier).hasPermission(MODIFY_PROPERTIES_PERMISSION);
+        } catch (RepositoryException e) {
```
delallioui, line 387: The catch here treats an access decision and a real store failure the same way — fail-closed `false`, DEBUG log, path suppressed. For the security verdict that's clearly right. But in the site-wide path a transient `RepositoryException` from one provider would now be indistinguishable from "caller may not write this node," and the only breadcrumb is at DEBUG. Would it be worth splitting the expected cases (`AccessDeniedException`/`PathNotFoundException` → debug, fail-closed) from an unexpected `RepositoryException` (log at WARN, still fail-closed)? I'm wondering whether an operator debugging "why did 30 nodes fail?" would otherwise have nothing to go on.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -292,6 +344,42 @@ static void flushNodeCaches(String path) {
...
+     * @throws DataFetchingException if the caller may not write the node's properties
+     */
+    private void ensureCallerCanModify(String nodeIdentifier) throws RepositoryException {
+        JCRSessionWrapper callerSession = JCRSessionFactory.getInstance().getCurrentUserSession(Constants.EDIT_WORKSPACE);
+        if (!isModifiableByCaller(callerSession, nodeIdentifier)) {
+            throw new DataFetchingException("Permission denied");
+        }
+    }
+
+    /**
```
delallioui, line 374: Just to confirm the intent captured in the javadoc: a node living only in `live` (published then deleted in edit) now keeps its tag on a site-wide rename/delete, where it used to be cleaned by the old sweep. That reads as the correct fail-closed trade-off to me — is there any cleanup path that used to rely on the old behaviour to purge orphaned `live` tags that we'd want to call out in the release/changelog?

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
```diff
@@ -300,8 +388,19 @@ private void validateNodeBelongsToSite(JCRSessionWrapper session, String nodeId,
...
+     * configured {@code TagHandler} on the value it stores, and that strategy is replaceable — a
+     * sanitizing implementation can reduce a non-blank argument to nothing, which would otherwise be
+     * written as an empty tag.
+     *
+     * @param newName the replacement tag value
+     * @throws GqlJcrWrongInputException if the value is blank, or normalizes to nothing
+     */
     private void ensureMutationTagName(String newName) {
-        if (StringUtils.isBlank(newName)) {
+        if (StringUtils.isBlank(newName) || StringUtils.isBlank(taggingService.getTagHandler().execute(newName))) {
```
baptistegrimaud: This normalizes the whole argument, but `TaggingService.renameTag` normalizes each **segment** after splitting on the tag separator — and on a stock install that separator is `,`, read from the `j:tagList` selector options.

- `TaggingService.renameTag` does `tagNewName.split(tagSeparator)` and then `newValues.add(tagHandler.execute(splitedTag))` per segment, so the value that lands in `j:tagList` is the *per-segment* normalization, not the whole-string one this guard inspects.
- `tagSeparator` is not null by default: it comes from `getPropertyDefinition("j:tagList").getSelectorOptions().get("separator")`, and the definition is `- j:tagList (string, tag[autocomplete=10, separator=','])`.
- So two inputs get past the guard and still land badly, with the stock `TagHandlerImpl` (`trim().toLowerCase()`):

| `newName` | this guard sees | what `renameTag` writes |
|---|---|---|
| `" , "` | `","` — non-blank, accepted | segments `[" ", " "]` → `execute(" ")` is `""` → an **empty tag value** in `j:tagList` |
| `","` | `","` — non-blank, accepted | segment array is empty → the old tag is dropped with **no replacement**, so a rename silently deletes |

- Matching the guard to what gets stored means checking the segments, which needs the separator — and `tagSeparator` is private on `TaggingService` with no getter, so the module would have to read the selector option itself (`NodeTypeRegistry.getInstance().getNodeType("jmix:tagged").getPropertyDefinition("j:tagList").getSelectorOptions()`, the same source core reads at init) or core would need to expose it.

**Not a regression — the whole-string form is what core checked too — but it is the one thing this guard was added to prevent, so it seems worth closing while you're here. Which way would you rather take it?**

eallais: Good catch, and the table is exactly right — I checked both shapes against `TaggingService.renameTag` (`splitedTags = tagNewName.split(tagSeparator)` then `newValues.add(tagHandler.execute(splitedTag))`) and against the selector option at `TaggingService`'s init.

Taken the first way you suggested: the guard now derives the stored values the way core does, in `c6ff9d2`. It reads the separator from the same source (`NodeTypeRegistry` → `jmix:tagged` → `j:tagList` → `selectorOptions`), splits, runs the tag handler per segment, and refuses a name whose segment list is empty or contains a blank. That keeps a legitimate multi-tag rename working, which rejecting the separator outright would have removed.

Two spec cases pin your two inputs: `" , "` and `","`. Verified on a 8.2.4.0-SNAPSHOT lab with the module deployed — 28/28.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
eallais: Fixed in `c6ff9d2`, though not by the exception type — that route does not discriminate here.

`JCRSessionWrapper.getNodeByUUID` walks every store provider and re-wraps whatever it caught as `ItemNotFoundException` (`throw new ItemNotFoundException(uuid, originalEx)`), so an `AccessDeniedException` from a provider and a broken session both arrive as the same type. What survives is the **cause**: absent or unreadable arrives with no cause or a missing-item one, a provider that broke arrives wrapped.

So the branch now logs at DEBUG when the cause is an expected absence and at WARN otherwise, fail-closed in both cases. Your operator debugging "why did 30 nodes fail?" gets a WARN per node for anything that is not an access decision.

Verified with the spec run: 28/28 on a 8.2.4.0-SNAPSHOT lab.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
eallais: Done in `c6ff9d2`.

Checked before switching, since the constant is the expanded form (`{http://www.jcp.org/jcr/1.0}modifyProperties`) rather than `jcr:modifyProperties`: `hasPermission` resolves through `JahiaPrivilegeRegistry.getPrivilege`, which calls `JCRContentUtils.getExpandedName`, and that method is idempotent on an already-expanded name (`if (!name.startsWith("{"))`). Both spellings land on the same key, so the constant is safe and the magic string is gone.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/service/tags/service/TagManagerMutationService.java
eallais: Confirmed, that is the intent — and no, nothing relied on the old sweep: across `jahia-private`, `graphql-core`, `siteSettings` and `jcontent`, the only callers of `renameTagUnderPath` / `deleteTagUnderPath` are these Tag Manager mutations, so there is no cleanup path losing a purge.

Called out in the changelog rather than in the javadoc alone: the fragment now carries the consequence, in `1f4bca4`. A node that exists only in `live` keeps its tag on a site-wide rename or delete, and is counted in `failedCount` without its path.

delallioui (approved):

PR #655, opened 2026-08-12
fix: added document limit to gql queries

r3dm1ke:
### Description
This PR adds a limit to the number of documents returned by a single GQL query.

#### Source code

#### Tests
- [ ] I've provided Unit and/or Integration Tests

> [!TIP]

### DIFF
--- a/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/relay/PaginationHelper.java
+++ b/graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/relay/PaginationHelper.java
@@ -21,6 +21,7 @@ import org.apache.commons.collections4.queue.CircularFifoQueue;
 import org.apache.commons.lang.mutable.MutableInt;
 import org.apache.commons.lang.mutable.MutableObject;
 import org.jahia.modules.graphql.provider.dxm.DataFetchingException;
+import org.jahia.modules.graphql.provider.dxm.GqlLimitExceededException;
 import org.jahia.modules.graphql.provider.dxm.node.GqlJcrWrongInputException;
 import org.jahia.modules.graphql.provider.dxm.util.StreamUtils;
 import org.slf4j.Logger;
@@ -36,6 +37,25 @@ public class PaginationHelper {
 
     private static Logger logger = org.slf4j.LoggerFactory.getLogger(PaginationHelper.class);
     private static AtomicInteger nodeLimit = new AtomicInteger(5000);
+
+    /**
+     * Default number of nodes one request may walk across every connection it selects, together.
+     * <p>
+     * Four times {@link #nodeLimit}, so a request can legitimately page through several connections in full, while a
+     * document that nests connections stays bounded: the per-connection limit caps each connection on its own, never
+     * their product, and it is the product that a nested selection multiplies out. Like the mutation batch limit this
+     * is a starting point rather than a measured figure - it is wide enough that observed traffic does not approach it,
+     * and is expected to be tightened once there is data on real request sizes.
+     */
+    public static final int DEFAULT_REQUEST_NODE_LIMIT = 20000;
+
+    private static AtomicInteger requestNodeLimit = new AtomicInteger(DEFAULT_REQUEST_NODE_LIMIT);
+
+    /**
+     * Key under which a request's remaining node allowance is held in the GraphQL context, seeded once per operation by
+     * {@code QueryCostInstrumentation} and drawn down by every connection the operation resolves.
+     */
+    public static final String REMAINING_NODE_ALLOWANCE = "jahiaRemainingNodeAllowance";
 
     private PaginationHelper() {
     }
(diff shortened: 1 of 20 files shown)


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/relay/PaginationHelper.java
```diff
@@ -179,6 +199,74 @@ public static int getNodeLimit() {
...
+        if (remaining == null) {
+            return stream;
+        }
+        if (remaining.get() < 0) {
+            // Already spent, and the connection that spent it has already said so.
+            return Stream.empty();
+        }
+        return stream.peek(node -> {
+            if (remaining.decrementAndGet() < 0) {
+                throw new GqlLimitExceededException("This request asked for more than " + getRequestNodeLimit()
```
gflores-jahia, line 252: Maybe also add a warn log whenever limit is bypassed

gflores-jahia (commented):
A few comments based on AI findings:

1. There are still some code that uses pagination (e.g. direct calls to `PaginationHelper.paginate()`) but doesn't seem to be covered by the `chargeToRequestBudget` i.e. does not use a stream version. 

Is this something that was intentionally skipped, or do we need to cover these scenarios as well? I copy-pasted findings based on `PaginationHelper.paginate()` references:

- **`TagManagerReadService.java:163`** — `tagManager(siteKey).taggedContent(tag)`. Worst finding of the audit: it uses the `List<T>` overload, which skips `collectItems` entirely — bypassing *both* the new request budget *and* the pre-existing per-connection `graphql.fields.node.limit` (5000) that's applied everywhere else. It returns real `GqlJcrNode` instances (via `SpecializedTypesHandler.getNode`), so `taggedContent(tag: "x") { descendants { nodes { descendants { ... } } } }` reopens the exact recursive-amplification vector this PR was written to close, fed by an unbounded, un-`LIMIT`'d JCR-SQL2 query.
- **`GqlGroup.java:112`** — `Group.members`. Self-recursive independent of `GqlJcrNode` entirely: a member can itself be a `Group`, which has its own `members` — `members { ... on Group { members { ... on Group { members { ... } } } } }` is structurally identical to the `descendants` bug. Uses the `Stream<T>` overload (the old 5000 cap applies per call), but each nested call draws a *fresh* 5000 rather than a shared budget — the same "per-connection cap bounds each level, never their product" flaw #655 exists to fix, just unfixed here. LDAP/AD-synced groups can hold tens of thousands of members.

** Lower severity (unbounded reads, but no `GqlJcrNode` recursion path):**

| Location | Field | Why lower severity |
|---|---|---|
| `GqlPrincipal.java:67` | `groupMembership` | Returns `GqlGroup` — chains back into the `members` bug above, but this call itself is bounded per-principal |
| `GqlUserAdmin.java:75` | `userAdmin.users` | `searchUsers(null)` — unbounded, all users across all providers, but `GqlUser` has no nesting field |
| `TagManagerReadService.java:115` | `tagManager(siteKey).tags(...)` | Same `List<T>`-overload bypass as `taggedContent`, but a plain tag DTO, not `GqlJcrNode` |
| `GqlJcrNodeImpl.java:419` | `node.references` | `GqlJcrProperty`, not nestable; unbounded JCR `PropertyIterator` |
| `GqlJcrNodeImpl.java:708` | `node.usages` | `GqlUsage` wraps a node but isn't itself nestable; unbounded |
| `GqlJcrNodeType.java:178,203,221` | `subTypes`/`extends`/`extendedBy` | In-memory `NodeTypeRegistry` reads — same category as the already-deliberately-exempted `getNodeTypes` |
| `GqlWorkflowService.java:86` | `activeWorkflowTasksForUser` | Plain POJOs, no JCR pivot, scoped to the calling user's own tasks |
| `GqlScheduler.java:78` | `paginatedJobs` | `List<T>`-overload bypass, but wraps Quartz jobs, not JCR content; small scale |
| `VanityUrlJCRSiteExtensions.java:127` | `getAllVanityURLs` | Same shape as the SDL finders (returns `GqlJcrNode`), but currently dead code — its `@GraphQLTypeExtension` is commented out, unreachable today |
| `SDLPaginatedDataConnectionFetcher.java:48` | all SDL finder fields (`allOffersConnection` etc.) | Returns `GqlJcrNode`; demonstrated concretely below |

2.  totalCount silently spends the full budget.

Could be as designed but could have major impacts on existing queries. Not sure if this has been spot tested against real workloads against full environment and flag for any potential impacts e.g. jcontent makes use a lot of totalCount across contents for the whole site for pagination. I can also provide a cypress test case to prove this if needed.

AI analysis below:

PaginationHelper.java:365-375 (StreamBasedDXPaginatedData.getTotalCount()) · DXRelay.java:62-68 (schema description) · PaginationHelper.java:252-254 (error message) · .chachalog/Nq4vZ7Rt.md:4 (changelog)

getTotalCount() drains the entire remaining match set to report an accurate count, completely independent of first/last/limit:
while (remaining.hasNext()) {
    remaining.next();
    totalCount++;
}
So descendants(first: 10) { nodes { name } pageInfo { totalCount } } against a 10,000-node subtree charges ~10,000 against the shared 20000-node budget — not 10. A completely ordinary "page 1 of N, show total count" UI pattern will now fail outright once the underlying content tree crosses the budget, where before this PR it just ran slow. Three places this should be surfaced:
- DXRelay.java:62 and :66 — both nodesCount and totalCount currently carry copy-pasted descriptions ("When paginating forwards, the cursor to continue.") that describe neither field. Nothing in the schema tells a consumer this field ignores pagination and can be this expensive.
- The GqlLimitExceededException message (PaginationHelper.java:252-254) tells a caller to fix this by "paginating with first/limit or nesting fewer levels" — neither helps when totalCount is what tripped it, since it ignores pagination entirely. Worth adding "or omitting totalCount, which reads the full remaining set" as a third remedy.
- The changelog note only says node reads are now bounded; it doesn't flag that totalCount usage is where an existing, non-malicious client is most likely to hit this for the first time. That's the kind of behavior-change callout the changelog style guide asks for.

(Side note, not actionable here: this incidentally closes a totalCount-based unbounded-read vector that existed independently of the nested-descendants CVE and predates this PR — the old per-connection graphql.fields.node.limit never gated getTotalCount()'s tail-drain, only page-building.)

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/relay/PaginationHelper.java
gflores-jahia: Unless I missed it,  we still don't have any server-side logging when this happens.

gflores-jahia (commented):
All has been addressed, except for logging

gflores-jahia (approved):

PR #658, opened 2026-08-21
chore(tests): bump graphql-dxm-provider test dependency to 3.9.0-SNAPSHOT

Fgerthoffert:
Part of Jahia/graphql-core#647

The 3.9.0 release [workflow run](https://github.com/Jahia/graphql-core/actions/runs/32489482012) failed while releasing the test modules: `misc-test-module` resolves `graphql-dxm-provider` through the `graphql-dxm-provider.version` property in `tests/jahia-module/pom.xml`, which still pinned `3.7.0-SNAPSHOT` — a snapshot that has since been purged from Nexus (it still existed when 3.8.x was released, which is why this never surfaced before).

This bumps the property to `3.9.0-SNAPSHOT`, the current development version published by main-branch CI.

Test-tree-only change — no product code affected, no changelog fragment.

### DIFF
--- a/tests/jahia-module/misc-test-module/pom.xml
+++ b/tests/jahia-module/misc-test-module/pom.xml
@@ -49,7 +49,7 @@
     <parent>
         <groupId>org.jahia.test</groupId>
         <artifactId>graphql-core-test-module-root</artifactId>
-        <version>3.9.0-SNAPSHOT</version>
+        <version>3.9.0</version>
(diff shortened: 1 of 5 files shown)


Fgerthoffert:
Superseded by #659, which fixes the same stale property by pinning the released `3.7.0` (which cannot expire from Nexus, unlike a snapshot). Closing without merge.

PR #663, opened 2026-08-26
fix: mutateNodesByQuery pages to its limit argument instead of refusing

baptistegrimaud:
Closes https://github.com/Jahia/jahia-private/issues/5420.

## Summary

`mutateNodesByQuery` treats its `limit` argument as a page size: it selects which slice of the matches to operate on, so a `limit` smaller than the number of nodes the query matched operates on that many instead of failing.

## Why

`limit` and `offset` on `mutateNodesByQuery` select a slice of the matches — "mutate the first N of them". [#645](https://github.com/Jahia/graphql-core/pull/645) introduced the mutation batch bound and folded the caller's `limit` into it as a single `effectiveLimit`, which the field then treats as a ceiling it refuses to go over. The two are not the same number, and merging them changed what `limit` means: `mutateNodesByQuery(query: "…", limit: 1)` against a query matching three nodes errors instead of mutating one.

Three things in #645 itself show that was not the intent:

- Its description says `mutateNodesByQuery` "is bounded when the query executes and **truncates** instead" — the merged code refuses.
- Its changelog fragment and the error message both point the caller at `limit`/`offset` on `mutateNodesByQuery` to work through a large result set — the same change made that advice fail.
- The comment on `aliasingAQueryDrivenMutationShouldNotMultiplyTheConfiguredMutationBatchLimit` reads "it truncates instead of being refused", while the assertion it guards relies on the whole request failing.

`GraphQLNodeMutationsTest.mutateNodesByQuery` is the test that catches it: it pages with `limit:1` against three matching nodes, then asserts the first of them carries the new title.

## Changes

- `GraphQLLimits.resolveMutationLimit(Long, AtomicInteger)` becomes `resolveMutationBatchBound(AtomicInteger)`: it resolves the *bound* only — the configured `graphql.mutation.batch.limit`, or what the request has left of it — and no longer takes the caller's argument. The single-argument overload, which only existed to serve it, is gone.
- `mutateNodesByQuery` compares the two rather than merging them. Within the bound, `limit` is a page and is applied as asked. Wider than the bound, the request is refused before the query runs, so whether it is allowed does not depend on how many nodes happen to match. With no `limit` there is nothing to compare yet, so that case stays bounded by the query and is checked as the result is read. Either way the bound draws from what the request has left, so aliased calls share one budget.
- Each refusal names what makes its number actionable: a field over the instance's own bound is told to operate on fewer at a time; one whose request has part of its allowance left is told how much, and to move the rest; one whose request has spent all of it is told to move the mutation to another request, because no page it asks for can fit.
- Regression tests for paging within the bound with and without `offset`, for a `limit` equal to the bound, for `limit: 0`, and for a `limit` wider than the bound with an `offset` past every match — which a check on the result set alone would let through.
- `aliasingAQueryDrivenMutationShouldNotMultiplyTheConfiguredMutationBatchLimit` gives each alias a `limit` within the bound, so it exercises the shared allowance rather than failing on the first alias, and asserts the error it expects.
- The default configuration file and `.chachalog/J8wXp.md` both described the bound as applying to what a request *operates on*; both now say it applies to what a request *asks for*, which is what the code enforces. `J8wXp.md` is amended rather than joined by a second fragment, since it is the fragment for this same behaviour.

## Validation

- `mvn compile` on `graphql-dxm-provider` and `graphql-test` (JDK 11).
- Every refusal message the tests assert was checked against the compiled `batchBoundExceeded` by reflection, so the test constants and the code agree on the exact wording.
- The integration suite runs in CI on this PR.

## Documentation

None needed — the `limit` argument's own description is already the paging one, and the configuration file's comment on the bound is updated with the change.

## ADR

None.


graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
```diff
@@ -178,13 +179,26 @@ public Collection<GqlJcrNodeMutation> mutateNodesByQuery(
...
         AtomicInteger remaining = remainingBatchAllowance(environment);
-        Long effectiveLimit = GraphQLLimits.resolveMutationLimit(limit, remaining);
+        Long bound = GraphQLLimits.resolveMutationBatchBound(remaining);
+        int configured = GraphQLLimits.getMutationBatchLimit();
+        Long page = (limit != null && limit.longValue() > 0) ? limit : null;
+        // A limit argument states how many nodes the caller wants to operate on, so asking for more than the bound is
+        // refused here, before the query runs: whether a request is allowed must not depend on how many nodes happen
+        // to match. Within the bound it is a page, and paging to it is what was asked for. See
+        // GraphQLLimits#resolveMutationBatchBound for why the two are compared rather than merged.
+        if (bound != null && page != null && page.longValue() > bound.longValue()) {
```
delallioui: This check compares the caller's `limit` against what the request has left of its allowance, and not against the configured bound.

`QueryCostInstrumentation` seeds that allowance as `maxBatchSize - cost.getBatchSize()`. `QueryCostCalculator` measures the enumerated batch items of the whole document, before any field runs. So `remaining` is already below the configured limit when a `mutateNodes` or an `addNodesBatch` field shares the document.

The effect is a refusal for a page size that the instance permits. Under a configured bound of 5000, the document `mutation { jcr { a: mutateNodesByQuery(query: "...", limit: 5000) b: mutateNodes(pathsOrIds: ["/x"]) } }` fails. Field `a` is refused, because `remaining` is 4999. Before this change `effectiveLimit` was `min(5000, 4999)`, and field `a` ran.

A `limit` equal to the bound is also the page size shown in `org.jahia.modules.graphql.provider-default.cfg` for a large result set. The recommended value is therefore the first one to be refused.

The pull request body gives an invariant that this check does not reach. The verdict no longer depends on how many nodes match, and it now depends on which other fields share the document.

One fix is to compare `page` against `GraphQLLimits.getMutationBatchLimit()`, and to keep `remaining` for what this field spends at line 221.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
```diff
@@ -209,6 +223,32 @@ public Collection<GqlJcrNodeMutation> mutateNodesByQuery(
...
+    private static String batchBoundExceeded(String lead, long bound, int configured) {
+        if (configured <= 0 || bound >= configured) {
+            return lead + "the maximum of " + bound + " it may operate on; operate on " + bound
+                    + " or fewer at a time, using the limit/offset arguments.";
+        }
+        if (bound <= 0) {
+            return lead + "what this request has left of its mutation batch allowance of " + configured
+                    + ", which earlier fields already used in full; move this mutation to another request.";
+        }
+        return lead + "the " + bound + " that earlier fields in this request left of its mutation batch allowance of "
```
delallioui: This message names "earlier fields in this request", and the number it reports also counts fields that come after the refused one.

`QueryCostInstrumentation` seeds the allowance as `maxBatchSize - cost.getBatchSize()`, and that cost measures the whole document before execution starts.

Under a bound of 5000, the document `mutation { jcr { a: mutateNodesByQuery(query: "...", limit: 5000) b: mutateNodes(pathsOrIds: ["/x"]) } }` refuses field `a` with "the 4999 that earlier fields in this request left". No earlier field exists, and field `b` has not run. An operator who reads that message looks for a preceding field that is not there.

graphql-dxm-provider/src/main/resources/META-INF/configurations/org.jahia.modules.graphql.provider-default.cfg
```diff
@@ -103,8 +103,10 @@ graphql.query.maxDepth = 30
 # Enumerated batches (mutateNodes, addNodesBatch) state their size in their arguments, so they are measured across the
 # whole document and refused before execution - which is also what stops the bound being multiplied by aliasing the
 # same mutation, or by burying items inside a recursive input type such as InputJCRNode.children. mutateNodesByQuery
-# cannot be measured that way, since how many nodes a JCR-SQL2 statement matches is not known until it runs, so it is
-# checked as it executes and draws from what the request has left rather than from the whole bound.
+# states its size the same way when it is given a limit, and is refused the same way, before the query runs; without
```
delallioui: This comment describes `mutateNodesByQuery` with a `limit` as a mutation that "states its size the same way" and "is refused the same way". The two mutations are refused against different numbers.

`GraphQLLimits.checkMutationBatchSize` refuses `mutateNodes` and `addNodesBatch` against the full configured limit. Line 189 of `GqlJcrMutation` refuses `mutateNodesByQuery(limit: N)` against what the request has left, and every other batch field in the document reduces that number.

An operator who sizes `graphql.mutation.batch.limit` from this comment expects `limit: 5000` to be accepted under a bound of 5000. Line 109 carries the same problem. A `limit` within the bound is a page across separate requests. Inside one request the second page is refused once the first page used the allowance.

graphql-test/src/main/java/org/jahia/test/graphql/GraphQLNodeMutationsTest.java
```diff
@@ -685,22 +725,57 @@ public void mutateNodesShouldRejectABatchLargerThanTheConfiguredMutationBatchLim
...
+        // offset picks which page, so the same limit of 1 mutates the second sub-list rather than the first.
+        withMutationBatchLimit(2, () -> {
+            validateNoErrors(mutateSubListTitlesByQuery(",limit:1,offset:1", PAGED));
+            assertTitledSubLists(PAGED, SUB_LIST_2);
+            return null;
+        });
+    }
+
     @Test
     public void aliasingAQueryDrivenMutationShouldNotMultiplyTheConfiguredMutationBatchLimit() throws Exception {
```
delallioui: No test puts `mutateNodesByQuery(limit: ...)` and an enumerated batch field in one document, and that is the combination where the new pre-query check changes behaviour.

Every new test calls `mutateNodesByQuery` alone or aliases it against itself, so `remaining` is always the full bound or 0. The branch of `batchBoundExceeded` for `0 < remaining < configured` runs in no test, and neither does the refusal at line 189.

One case covers both. Add `mutation { jcr { q: mutateNodesByQuery(query: "...", limit: 2) n: mutateNodes(pathsOrIds: ["/testList/testSubList3"]) } }` under `withMutationBatchLimit(3, ...)`.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
```diff
@@ -158,10 +158,11 @@ public Collection<GqlJcrNodeMutation> mutateNodes(
      *
      * @param query         the query to retrieve the nodes to be modified
      * @param queryLanguage the query language
-     * @param limit         the maximum size of the result set; the request fails if more nodes match
+     * @param limit         the maximum size of the result set
```
delallioui: The client-visible description of `limit` at line 172 still shows "The maximum size of the result set". This `@param` line also drops the sentence that warned about the bound.

Before this change the `@param` gave "the maximum size of the result set; the request fails if more nodes match". The argument now fails the request before the query runs when it exceeds the instance bound. Line 2131 of `tests/schema.graphql` shows the `@GraphQLDescription` unchanged.

A client that introspects the schema reads a plain maximum, and it picks `limit: 10000` on an instance whose `graphql.mutation.batch.limit` is 5000. The schema states no upper bound for the argument.

Under Documentation the pull request body gives this reason: "the `limit` argument's own description is already the paging one". That description names no page and no bound.

.chachalog/xuihGDnM.md
```diff
@@ -0,0 +1,7 @@
+---
+graphql-core: patch
+---
+
+Fixed paging on `mutateNodesByQuery`: a `limit` smaller than the number of nodes the query matches now operates on that many nodes instead of failing the request. This is the paging that the 3.9.0 note on `graphql.mutation.batch.limit` points to for splitting a large result set across several requests.
+
+A `limit` wider than that bound is still refused, and is now refused whatever the query matches, so the outcome no longer depends on how much content happens to be there.
```
delallioui: Two sentences in this fragment pass the 25-word ceiling, and "that bound" names a bound that the fragment never introduces.

`style_check.py` reports two `long-sentence` violations, at 26 words and at 31 words. The first paragraph names only "the number of nodes the query matches". The second paragraph then shows "A `limit` wider than that bound is still refused", so a reader looks backwards for a bound and finds none.

The second paragraph can name `graphql.mutation.batch.limit` instead.

delallioui (commented):
Six comments from an independent review pass on this branch. The comment on `GqlJcrMutation.java` line 189 reports a behaviour change against version 3.9.0. The other five cover a refusal message, the operator comment on the bound, the schema text for `limit`, and a gap in the tests.

This pass read the branch and the repository only. It read nothing about the issue behind the change, so it says nothing about the acceptance criteria. It ran no build and no test.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
baptistegrimaud: Addressed in [`f2439b4a`](https://github.com/Jahia/graphql-core/pull/663/commits/f2439b4a7d8a296ed003dcfd1bbc55df1531fcc4) — both messages now say "other fields", with the reason recorded in `batchBoundExceeded`'s javadoc so it does not come back.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
baptistegrimaud: Addressed in [`f2439b4a`](https://github.com/Jahia/graphql-core/pull/663/commits/f2439b4a7d8a296ed003dcfd1bbc55df1531fcc4).

The description now states what the argument selects and that a value above the bound is refused, and the `@param` names both refusal causes. `tests/schema.graphql` is regenerated from a live instance with the module deployed — note that also sweeps in description drift left by earlier changes, so that file's diff is wider than this one.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
baptistegrimaud: Addressed in [`79c8157e`](https://github.com/Jahia/graphql-core/pull/663/commits/79c8157e77909f691a000ded478956bfc1592421).

Confirmed, and it was a regression against 3.9.0: `QueryCostCalculator.calculate` walks the whole document before execution, so an enumerated sibling lowers the allowance before any field runs. Your example is exact.

I did not take the suggested fix literally, though. Comparing `page` against `getMutationBatchLimit()` and keeping `remaining` only for the spend would let aliased query-driven calls each pass the check, which is the multiplication the bound exists to prevent. So the argument is measured against the configured limit, and the allowance still bounds what the field actually takes, settled as the result is read. On your document there is no pre-refusal and the outcome matches 3.9.0.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
baptistegrimaud: Addressed in [`79c8157e`](https://github.com/Jahia/graphql-core/pull/663/commits/79c8157e77909f691a000ded478956bfc1592421) — both messages now say "other fields", with the reason recorded in `batchBoundExceeded`'s javadoc so it does not come back.

graphql-dxm-provider/src/main/resources/META-INF/configurations/org.jahia.modules.graphql.provider-default.cfg
baptistegrimaud: Addressed in [`79c8157e`](https://github.com/Jahia/graphql-core/pull/663/commits/79c8157e77909f691a000ded478956bfc1592421).

Most of it follows from the `limit` check above: the argument is now refused against the same value as an enumerated batch, which is what this comment claimed. The block also separates whether a limit is legal from whether one request has the room for it, and says where each of the two forms is measured.

A later pass corrected more of this block than your comment asked for: it claimed the shared budget is drawn on by the fields named at the top, when in fact every argument that hands a mutation nodes to write draws on it — `addNode`/`addChild` children, `addChildrenBatch`, `copyNodes`, `moveNodes`, `mutateVanityUrls` included.

graphql-test/src/main/java/org/jahia/test/graphql/GraphQLNodeMutationsTest.java
baptistegrimaud: Addressed in [`e77d08c5`](https://github.com/Jahia/graphql-core/pull/663/commits/e77d08c541afa89355cfe11dd7f8e1fc27bc0d39).

Two tests put a query-driven page beside an enumerated batch; one of them fails on the previous code, so it is a regression test rather than only coverage. Following the thread turned up three more gaps, covered in the same commit: a page against a disabled bound, a non-positive limit, and `validateNoErrors` on the tests that assert repository state — including `mutateNodesByQuery` itself, which is why the original failure showed a missing title rather than the refusal that explained it.

graphql-dxm-provider/src/main/java/org/jahia/modules/graphql/provider/dxm/node/GqlJcrMutation.java
baptistegrimaud: Addressed in [`79c8157e`](https://github.com/Jahia/graphql-core/pull/663/commits/79c8157e77909f691a000ded478956bfc1592421).

The description now states what the argument selects and both ways a value can be refused — above the configured limit, or with the room already claimed by the rest of the request — and the `@param` and `@throws` say the same. `tests/schema.graphql` is regenerated from a live instance with the module deployed — note that also sweeps in description drift left by earlier changes, so that file's diff is wider than this one.

.chachalog/xuihGDnM.md
baptistegrimaud: Addressed in https://github.com/Jahia/graphql-core/pull/663/commits/79c8157e77909f691a000ded478956bfc1592421.
_Note that this PR is now based on [#665](https://github.com/Jahia/graphql-core/pull/665), which reverts the 3.9.0 release, so `.chachalog/J8wXp.md` got restored._

delallioui (approved):

